PR #6, opened 2017-11-27
Feature/export protocol

wwqrd:
First run of export feature.

Serialises protocol JSON into file, as well as packaging theoretical assets from `protocol.assetRegistry`.


jthrilly:
I'll try and check this out today, but if not I will look at it first thing tomorrow.

src/utils/exporter.js
```diff
@@ -0,0 +1,80 @@
+import { remote } from 'electron';
+import fs from 'fs';
+import Zip from 'jszip';
+import { has } from 'lodash';
+
+const options = {
```
wwqrd: should be more specific

src/utils/electron-shim.js
```diff
@@ -0,0 +1,9 @@
+/* eslint-disable no-console */
+
+module.exports = {
+  remote: {
+    dialog: {
+      showSaveDialog: (_, callback) => { callback('demo.netcanvas'); },
```
wwqrd: `[...].canvas` is the filename convention

PR #7, opened 2017-11-29
Feature/refactor timeline components for reuse

wwqrd:
I extracted a couple of things into components / HOCs to make developing EditStage simpler.

Also fixed the Zoom HOC to scale properly, and extracted constraints out into context using recompose (meaning it's an easy change to revert).


jthrilly:
The animation is perfect now!

PR #24, opened 2018-03-09
Fix/exporter

wwqrd:
- Fixes the export to respect `assets/` folder.
- Updates to Information Interface to match protocol spec.

# Known issues

If Information interface has no items, Network Canvas will not load the stage. We're going to update NC to make `items` an optional key.

Based off feature/edit-stage, so that should probably merge first, this branch starts at: bfba722


wwqrd:
For some reason this won't build on travis, but it builds locally. I'm going to go ahead and merge this and revisit that later.

PR #30, opened 2018-03-28
Edit Name Generator (Work in progress)

wwqrd:
Adds features to enable editing a Name Generator Stage in a Network Canvas Protocol.

Functionally working, and have revisited Instructions edit components to bring them up-to-date with redux form components.

Know issues:
- Lacks animation, including:
  - [ ] Adding/removing variables from prompts
- [ ] Lacks error styling and display
- Should share form selector by passing down props
- Edit stage classnames / folder structure could do with a tidy up
- Missing component tests.


jthrilly:
Trying to take a look at this, but not able to actually save what I create (presumably because of not being able to select a node type?)

jthrilly:
Also, I can't edit and save an information interface either. When I click to save it, I get "failed". The generated code looks good.

One last thing - it looks like if you open a protocol that has existing information interfaces (such as the teaching factory protocol) you can't see/edit the content items. The name generators load though, which is really cool.

PR #44, opened 2018-04-25
Component tweening WIP

wwqrd:
Here's my attempt at component tweening. I'm submitting this as a work in progress pull request, but it would be great to get your thoughts on it.

It creates a clone (outside of React), of the `from` element, and then applies a transform to that to match the `to` element. It adds some predictably named styles to allow for component specific animations.

Tweens exist in a global object and can be called from anywhere in the application, in this case they are called in a route change listener.

Good things:
- It works! (Wasn't sure this would be possible without going full jQuery...)
- Lots of flexibility for what we can animate

Bad things:
- Complexity (but I haven't thought of better). We have css animations between routes, js tweens between transform states, and additional css transitions on the individual component. To orchestrate animations between routes, all three must be considered.
- JS tweens are absolute, from top right corner, and do not respect scrolling.
- `window.Tweens`?
- Using `delay()` (equiv to `setTimeout( , 1)`), to ensure both elements are on the stage before the animation starts.
- The routes dispatcher seems a bit hacky - over all none of this feels like The React Way

# How to see the end result

1. Open up the app in webpack preview or electron:

    `npm run start`

1. Click one of the recent protocol on the start screen (if there isn't one already click "create new")
1. You should see it "tween" to the overview card on the next screen.

# Known issues

This demo uses "foo" as the name for all animations, so if there is more than one protocol in the recent items list on the start screen it may behave strangely.


wwqrd:
FYI: there are some bad property names I'm planning to change:
`name` -> `tweenName`
`uuid` -> `tweenTargetName`

src/behaviours/Tweened/index.js
```diff
@@ -0,0 +1,82 @@
...
+    componentWillUnmount() {
+      this.node.removeEventListener('click', this.onClick);
+
+      window.Tweens = {
+        ...window.Tweens,
+        [this.props.name]: omit(window.Tweens[this.props.name], this.uuid),
+      };
+    }
+
+    onClick = () => {
```
bryfox: Is this legacy (replaced by routing tweens)? I don't see a listener added anywhere.

src/routes.js
```diff
@@ -1,32 +1,77 @@
...
+class Routes extends Component {
+  componentDidMount() {
+    this.props.history.listen(dispatchRouteAnimations);
+  }
+
+  render() {
+    const { location } = this.props;
+    return (
+      <TransitionGroup component={null}>
+        <CSSTransition appear timeout={getCSSVariableAsNumber('--animation-duration-fast-ms')} classNames="route" key={location.key}>
```
bryfox: So this means that all screen transitions/tweens need to have the same duration (and have a tween, period)? That's fine, if that's the design goal. Just curious...

src/behaviours/Tweened/index.js
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    componentDidMount() {
+      this.node = ReactDOM.findDOMNode(this); // eslint-disable-line react/no-find-dom-node
+
+      window.Tweens = {
+        ...window.Tweens,
+        [this.props.name]: {
+          ...window.Tweens[this.props.name],
+          [this.uuid]: {
```
bryfox: This looks like it's called at the point of adding a tweened destination. e.g.,  'overview-panel' is added when I click the 'protocol-stack'. But its position info doesn't change before the cached information runs. So if the position changes between route changes, the tween starts from the wrong place.

Steps:
- Click protocol stack
- Change browser viewport width to 50% or so
- Click overview component

Seen: tween starts from the old (cached) position/dimensions of the overview component.

bryfox: I don't mean to nitpick on the tweening behavior; I'm more trying to understand how the tween objects are initialized and cached (and determine if they need to be.)

src/behaviours/Tweened/index.js
```diff
@@ -0,0 +1,82 @@
+import React, { PureComponent } from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import { omit, pick } from 'lodash';
+import uuid from 'uuid';
+import getAbsoluteBoundingRect from '../../utils/getAbsoluteBoundingRect';
+import tween from './tween';
+
+window.Tweens = window.Tweens || [];
```
bryfox: Since this is all 'out of' react anyway, limited use of window seems fine to me. But if it's bothering you, would a separate redux store work? (Would help debugging, if nothing else.)

src/routes.js
```diff
@@ -1,32 +1,77 @@
...
-        />
-        <Redirect to="/" />
-      </Switch>
-    </CSSTransition>
-  </TransitionGroup>
-);
+const dispatchRouteAnimations = ({ pathname }) => {
+  switch (true) {
+    case /^\/$/.test(pathname):
+      tween({ name: "foo", from: "overview-panel", to: "protocol-stack" });
```
bryfox: The tween back to the protocol stack is interesting, since the click area includes the titles below (and margin around) the image, but you'd probably just want to tween to the image. Is this flexible enough to delegate a different view object to transition or does that throw a wrench in things?

src/behaviours/Tweened/index.js
wwqrd: Correct!

src/routes.js
wwqrd: That's an interesting point: we have references to both from and to elements, we could clone them both!

Probably the way I'd implement that is that the overview panel would fade to a solid color as it moves over the top of where the protocol stack will appear, once there it fades out to reveal the stack underneath, which could fade in titles at the same time - this is part of the issue I have with this implementation, all the synchrony required.

src/routes.js
wwqrd: Because we are using cloned elements for the tween they *could* differ, though more than likely we won't want that. Tween takes an argument for durations, so I should probably update this so that they relate somehow...

src/behaviours/Tweened/index.js
wwqrd: Hmm, I think I will try only registering the elements here, and then getting the positions on tween.

Also, using a redux store for `window.Tweens`, like you suggested above, might make the intent clearer.

wwqrd:
Merged in in timeline branch

PR #46, opened 2018-05-09
Feature/new icon

jthrilly:
Implements a new app icon, removes some Cordova cruft, and puts the package version at 4.0.0-alpha.1 in line with everything else.


wwqrd:
Looks good, particularly like the 'junk' removal 😅

PR #47, opened 2018-05-10
New stage panel enhancement

wwqrd:
This adds a draw style new stage panel, which clarifies the UI spatially.


wwqrd:
@bryfox Added you to this PR in case it interests you.

PR #51, opened 2018-05-14
Add basic electron-dev workflow

bryfox:
Feel free to change as needed; this is basically what I've been using in Server, and it meets my needs.

I tested quickly; it seems to run OK in both forms (web/browser & electron), but let me know if any functionality has broken.

Startup commands are the same as Network Canvas: `npm run start:electron` for renderer, and `npm run electron:dev` for main. However (like Server), this is a simpler version that assumes you're running frontend on port 3000.


package.json
```diff
@@ -48,6 +51,7 @@
         "classnames": "^2.2.5",
         "color": "^2.0.0",
         "connect-history-api-fallback": "1.3.0",
+        "cross-env": "^5.1.5",
```
bryfox: Takes care of Windows issues with env vars.

.gitignore
```diff
@@ -9,6 +9,9 @@
 /docs-build
 /plugins
 
+# local dev build output
+electron-dev/
```
bryfox: Everything runs out of this dir for development; in this case, it's basically a copy of `public/`.

`www` is left alone as the existing build output dir.

public/electron-starter.js
```diff
@@ -31,17 +31,28 @@ log.info('App starting...');
...
 
   // Create the browser window.
   mainWindow = new BrowserWindow(windowParameters);
   mainWindow.maximize();
   // and load the index.html of the app.
-  mainWindow.loadURL(url.format({
-    pathname: path.join(__dirname, 'index.html'),
-    protocol: 'file:'
-  }));
+  mainWindow.loadURL(getAppUrl());
```
bryfox: This is the main runtime change.

package.json
```diff
@@ -11,7 +11,10 @@
     },
     "scripts": {
         "pack": "electron-builder --dir",
-        "start": "node scripts/start.js",
+        "start": "cross-env BUILD_TARGET=web node scripts/start.js",
```
bryfox: To support browser workflow, I added the `BUILD_TARGET` flag. Depending on your workflow, you might want to flip this on its head so that `web` is default and `electron-renderer` is the special case. (On Server, I always run with electron.)

wwqrd:
Looks ideal, thanks!

PR #52, opened 2018-05-16
Remove duplicate copy task

bryfox:
This is a minor update to the electron:dev build script, which just popped into my head.

When I adapted the electron-dev script from server, I left in an extra copy step that isn't needed; public is already copied. (Server has its main JS outside the webpack bundle, so it uses the 'mainSrc' copy step.)


wwqrd:
👍

PR #53, opened 2018-05-16
Feature/open dot netcanvas files

### LINKED ISSUE #45: Work with bundled versions of protocols by default, and project folders as a 'hidden' advanced feature.
Users should be able to open `.netcanvas` bundles, and when saving changes these would go directly into the bundle (without an 'export' step).

Advanced users should be able to open a `protocol.json` file in a project folder, this will reveal additional UI for exporting bundles.

wwqrd:
This switches from working with project directories to "directly" editing the archive format.

Underlying changes have preserved the fact that behind the scenes we are still working with project directories (in /tmp), so that we should easily be able to reimplement the "advanced" functionality of working with project directories later down the line.

Archives are currently uncompressed for speed. Given that assets will probably already be compressed, I wouldn't expect this to have a big impact on file size - but it's easy enough to change.

resolves #45


wwqrd:
@rebeccamadsen Looks like I broke something 😬 Thanks!

wwqrd:
Okay, that should work now 👍

PR #54, opened 2018-05-17
implement .netcanvas file association for editor role

jthrilly:
This PR implements the `.netcanvas` file association, and assigns Architect the editor role. It also adds a file icon, and updates the build scripts to use it.

To test this, run `npm run dist:[platform]` and then install the resulting app. On macOS, drag the `.app` to the Applications folder.


wwqrd:
Looks good to me!

Issue #42, opened 2018-04-23
Architect architecture


wwqrd:
We are using submodule architecture for most components, but there are still remnants of the earlier container/component structure. Update these components to be consistent. One exception might be the "Screen" containers, perhaps these could have their own directory.

wwqrd:
Resolved

PR #59, opened 2018-05-17
Update to use a nested component structure

wwqrd:
Updates previously established container/component components to use
a nested structure as defined here: https://github.com/codaco/Architect/wiki/Component-Organisation


wwqrd:
Fixes #42

PR #61, opened 2018-05-17
Use system font stack

### LINKED ISSUE #19: System font stack
Our desktop apps should use the system font of the host platform for body text, reserving our Network Canvas font stack (Quicksand) for headings, and marquee text.

wwqrd:
Resolves #19


wwqrd:
@jthrilly Actually think this looks a bit odd with the smaller NC headings, wonder if we should just use the NC font for page/section headings?

jthrilly:
It might be a case of tweaking the smaller heading styles. Let me take a look.

jthrilly:
It looks to me like increasing the heading weight would work as a fix for this (500 seems quite good). The quicksand heading weight is lighter than the body type of most of the system fonts, and I think the heading weight is perhaps a bit light anyway, particularly for desktop.

If this becomes more of a pain, I think your suggestion of reserving Quicksand for important headings would work.

jthrilly:
One thing I noticed is that buttons and similar UI elements should still be Quicksand, I think. Perhaps we should specify this in UI?

wwqrd:
This just sets the font for body, so if it's not showing up on buttons I don't think the font has been specified for those in UI, will do a PR there.

PR #63, opened 2018-05-22
Feature/open from file system

### LINKED ISSUE #55: Handle opening .netcanvas files by file association
Pr #54 adds a file association for Architect and .netcanvas files.

We now need to handle these files within Architect. From what I can tell, the file gets passed as an argument when the app is launched, but to handle this correctly we need to ensure we are running as a single instance: https://github.com/electron/electron/issues/8006

Once that is done, we can look to `argv` to get the path to the opened file. This will then need to be passed to the renderer side, and then used to trigger the file open methods.

wwqrd:
This builds on the file association work to allow file opening from clicking on a protocol file! :tada:

It also adds a menu `file > open` option taken from Server.

Windows support is complete theoretical, so would be great if you could check it out @rebeccamadsen? I believe you need to build and then install it for the file association stuff to take effect.

### Running the app (file > open)
The web version of the app needs some more "shim work", so to run it in dev you're better off with the live reload electron setup:
```bash
$ npm run start:electron 
$ npm run electron:dev
```

### Running the app (file association)
```bash
$ npm run dist:[os]
# Follow installation steps for OS
```

Resolves #55


bryfox:
Based on #54, it sounds like I need use a distribution build testing click-to-open. When I run `npm run dist`, I get the following version error:

```
Error: At least electron-updater 2.18.2 is required by current electron-builder version. Please set electron-updater version to "^2.18.2"
```

rebeccamadsen:
I don't have certificate, so I can't build the .exe, but I can test one if you want to send it to me. Or if you tell me how to get a certificate, either way.

```
Error: Cannot find certificate Complex Data Collective, all certs:
    at C:\Users\Big Baloo\Documents\GitHub\Architect\node_modules\electron-builder-lib\src\windowsCodeSign.ts:114:12
    at Generator.next (<anonymous>)
    at runCallback (timers.js:794:20)
    at tryOnImmediate (timers.js:752:5)
    at processImmediate [as _immediateCallback] (timers.js:729:5)
From previous event:
    at getCertificateFromStoreInfo (C:\Users\Big Baloo\Documents\GitHub\Architect\node_modules\electron-builder-lib\src\windowsCodeSign.ts:123:2)
    at vm.value.then.vm (C:\Users\Big Baloo\Documents\GitHub\Architect\node_modules\electron-builder-lib\src\winPackager.ts:30:47)
    at <anonymous>
```

The dev version is okay for me on windows -- I can open .netcanvas files using the "Open Existing" button and File->Open. Although I get this on console when I start up:
```
Uncaught (in promise) Error: ENOENT: no such file or directory, open 'C:\Users\Big Baloo\Documents\GitHub\Architect\electron-dev\protocol.json'
    at Object.fs.openSync (fs.js:652:18)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:172:20)
    at Object.fs.readFileSync (fs.js:553:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:508:29)
    at loadProtocolData (http://localhost:3000/static/js/bundle.js:138189:64)
    at http://localhost:3000/static/js/bundle.js:137028:106
    at http://localhost:3000/static/js/bundle.js:105381:18
    at dispatch (http://localhost:3000/static/js/bundle.js:105799:18)
    at http://localhost:3000/static/js/bundle.js:137043:7
    at <anonymous>
```

And a couple of unique key prop warnings on console.

wwqrd:
@rebeccamadsen I can't sign it either, the cert is a physical key! Is there not a way to do it without signing?

That fs.openSync error looks like the file open stuff for windows isn't working at the moment anyway 👍

wwqrd:
@rebeccamadsen Managed to do an (unsigned) build on mac by removing the certificateSubjectName, does that work for you?:

```diff
diff --git a/package.json b/package.json
index 76b32a4..c318b8b 100644
--- a/package.json
+++ b/package.json
@@ -159,8 +159,7 @@
             "category": "public.app-category.education"
         },
         "win": {
-            "target": "nsis",
-            "certificateSubjectName": "Complex Data Collective"
+            "target": "nsis"
         },
         "fileAssociations": {
             "ext": "netcanvas",
```

Tested app on Parallels (vm), and it seems to open the protocol correctly in Architect with the most recent changes :tada:

rebeccamadsen:
Yes! Removing the certificateSubjectName worked for me! As far as I can tell things are working here. I don't see any of those errors from before, I was able to create and install the .exe and there are cute little icons for the .netcanvas files now. And clicking on .netcanvas files runs and opens Architect, so yay! 

If I already have Architect running, I end up with two running Architects. Not sure if that is desirable? I'm curious what happens if you end up trying to edit the same file in both instances, but have not tried that.

Let me know if there is anything else you'd like me to try.

wwqrd:
@rebeccamadsen I've written up a separate issue for the multiple window behaviour here: #65

PR #64, opened 2018-05-22
Added 'Develop' menu option

### LINKED ISSUE #56: App menu utilities for alpha
Recent changes to the app (redux session) were crashing it in dist, without devTools, there isn't a simple way to recover.

It'd be great to:
- Have a quick way to reset the app from the menu
- Open dev tools with a keyboard shortcut.

wwqrd:
Adds an extra menu option to completely resolve #56

### DIFF
--- a/public/components/mainMenu.js
+++ b/public/components/mainMenu.js
@@ -1,4 +1,4 @@
-const { dialog } = require('electron');
+const { dialog, session } = require('electron');
 
 const openDialogOptions = {
   buttonLabel: 'Open',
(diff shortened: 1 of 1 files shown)


public/components/mainMenu.js
```diff
@@ -45,17 +45,36 @@ const MenuTemplate = (window) => {
...
+        { type: 'separator' },
+        {
+          label: 'Clear storage data',
+          click: () => {
+            dialog.showMessageBox({
+              message: 'This will reset all app data, are you sure?',
+              buttons: ['OK', 'Cancel'],
+            }, (response) => {
+              if (response === 0) {
+                window.webContents.session.clearStorageData();
```
bryfox: I just noticed that [this is async & takes a callback](https://electronjs.org/docs/api/session#sesclearstoragedataoptions-callback). This seems to work as-is, but might want to move reload into the callback to be safe?

wwqrd: Good spot, thanks!

PR #66, opened 2018-05-25
Feature/edit overview panel

### LINKED ISSUE #49: Protocol overview panel needs to be implemented
This has a few deeper dependencies which make it difficult to implement at the moment:
- Variable registry
- Form editor
- Asset browser

Suggest we start with the "global options" section, and add these other dependencies as we add them.

wwqrd:
Basic overview panel showing some information about the protocol and allows protocol title to be edited.

Resolves #49

### DIFF
--- a/src/components/Timeline/Overview/Overview.js
+++ b/src/components/Timeline/Overview/Overview.js
@@ -7,32 +7,37 @@ import PropTypes from 'prop-types';
 import { get, keys } from 'lodash';
 import { map, groupBy, toPairs } from 'lodash/fp';
 import { compose } from 'recompose';
-import { Image } from '../Assets';
-import { Contexts, SeamlessText } from '../Form/Fields';
-import { Tweened } from '../../behaviours/Tweened';
-import { actionCreators as protocolActions } from '../../ducks/modules/protocol';
-
-const PanelGroup = ({ title, children }) => (
-  <div className="timeline-overview__group">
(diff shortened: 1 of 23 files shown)


src/components/Timeline/Overview/Overview.js
```diff
@@ -0,0 +1,143 @@
...
+    const assets = this.props.assets;
+
+    return sortAssets(assets)
+      .map(([groupName, group]) => {
+        if (!['.png', '.jpg', '.gif'].includes(groupName)) {
+          return <div className="timeline-overview__badge"><em>{groupName}</em> x {group.length}</div>;
+        }
+
+        return (
+          <div className="timeline-overview__asset-group">
```
bryfox: I'm getting warnings about missing `key`s here, also :52 and probably :45.

src/components/Timeline/Overview/Overview.js
```diff
@@ -0,0 +1,143 @@
...
+  groupBy('extension'),
+  map(assetPath => ({ assetPath, extension: path.extname(assetPath) })),
+);
+
+class Overview extends Component {
+  get renderNodeTypes() {
+    const nodeTypes = keys(get(this.props.variableRegistry, 'node', {}));
+
+    return nodeTypes.length > 0 ?
+      <Contexts options={nodeTypes} input={{ value: null, onChange: () => {} }} /> :
```
bryfox: PropTypes warnings here: missing some required props.

Issue #65, opened 2018-05-24
Opening .netcanvas files behaves inconsistently across platforms


wwqrd:
Currently:

OSX:
Single window, when opening a file, currently opened file is "closed"

Windows:
Multiple windows, can open the same .netcanvas file in multiple windows simultaneously.

Proposed behaviour (both platforms):

When opening different .netcanvas files, open new windows, when opening the same .netcanvas file, focus window.

wwqrd:
Partly fixed, in that both platforms now act like OS X. Plan to revisit this.

wwqrd:
@jthrilly What's the intended behaviour for this?

jthrilly:
I think just something "deliberate". Having the app be single window is fine, but to finish this feature off we should check for things like "closing the current file (to open a new one) would cause data loss". If that's not an issue, I'm happy with how it is working now.

wwqrd:
I think given that we have a home screen that allows the user to switch between protocols from within the app - the simplest thing to do right now is keep this as a single window application.

So for this task I think the resolution is:
- Fix the multiple window shortcut from #164 
- Introduce a prompt to save when application is about to switch protocols.

PR #73, opened 2018-05-29
Force single app in electron

wwqrd:
- Force single app in electron
- Refactor app/window creation to make this easier to manage.

Temporary resolution to: #65

### DIFF
--- a/public/components/appManager.js
+++ b/public/components/appManager.js
@@ -1,0 +1,55 @@
+const { ipcMain, app } = require('electron');
+const path = require('path');
+const windowManager = require('./windowManager');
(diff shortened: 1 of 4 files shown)


wwqrd:
After speaking with @jthrilly we decided to make both os x and windows behave the same as os x default for now: opening each file in the original window.

rebeccamadsen:
Looks good to me! If the app is open, it opens the file I clicked on. If I had a different file open, it only shows the window for the newest one I clicked on. It reloads the file if I click on a file that I already had open. Let me know if you want me to try anything else!

PR #106, opened 2018-06-13
Refactor guidance component

wwqrd:
This changes the guidance components so that we can improve the
animations and better organise the guidance default copy.


wwqrd:
@jthrilly This is built off the Sociogram interface work, so really should be merged into that branch and not master.

Please review look and feel & take a look how the content is no organised, but don't merge it!

jthrilly:
This looks great! Big improvement. Feedback:

- Can the subcomponents of the prompt on the sociogram config have individual guidance? I guess the broader question: is the guidance limited to a specific "level"?
- I know it won't perform well, but can we animate the editable portion's width, to help with guidance expanding/collapsing? Something like this on `.guided__content` looked okay (obviously replace with NC animation vars):     
```
transition: width 0.5s ease-in-out;
transition-delay: 0.3s;
```

wwqrd:
> - Can the subcomponents of the prompt on the sociogram config have individual guidance? I guess the broader question: is the guidance limited to a specific "level"?

We're in a much better place to make something like that happen now, but it would require some changes - I think we'd need to make the component redux driven. Want me to go ahead and do that?

> - I know it won't perform well, but can we animate the editable portion's width, to help with guidance expanding/collapsing? Something like this on .guided__content looked okay

👍

jthrilly:
> We're in a much better place to make something like that happen now, but it would require some changes - I think we'd need to make the component redux driven. Want me to go ahead and do that?

I'm not sure, honestly. It just struck me that each "prompt" has quite a few subcomponents to configure on this interface, and that might be a pattern we find on other interfaces, too. What is your view? I sort of hoped it would just be a matter of creating more guidance components and it would work automagically (since elsewhere we have guidance "per input"), but I guess it is more complicated than that.

wwqrd:
At the moment it's a guidance component per-section. I'd imagined that the guidance section would be quite long in that case of SociogramPrompt, with subheadings.

One thing we could do is some kind of "scroll to" effect - I think it'd look great, but I'm not sure it's good UI.

What would you think about a system like:
```jsx
<Guided  /* This component triggers the "on hover" action */
  guidance={() => (<p> guidance goes here, awesome!</p>)}
>
  <input />
</Guided>```

I think a redux based implementation (like above), would work best based on your feedback.

We can keep the display component from Guidance v2 but it sounds like a more flexible dispatch guidance from anywhere approach might work better for lots of little component guidance sections.

It could also be used throughout the app to drive any other guidance interfaces we wanted.

jthrilly:
I think we discussed the above on Slack, and decided to go with it. Looks like a good approach to me!

PR #111, opened 2018-06-25
Sociogram interface with improved guidance component

### LINKED ISSUE #14: Sociogram configuration interface

wwqrd:
Sorry this is such a huge PR, it's a couple of inter-woven features which led to other smaller additions/changes.

__Requirements: will need to update UI reference to latest master to test__

Resolves #14

# Sociogram Interface

screen shot 2018-06-25 at 16 49 31

Adds new edit capabilities for a Sociogram Interface, mainly focused around implementing the SociogramPrompt (https://github.com/codaco/Architect/compare/master...feature/sociogram-interface-with-guidance#diff-f5e8507d1b1855332efaecf5ed24ce6d)

Related additions:
  - new `OrderBy` field component (`<Field component={OrderBy}>`), which is a mini sortable component much like the prompts (https://github.com/codaco/Architect/compare/master...feature/sociogram-interface-with-guidance#diff-2432403a3a9045686924ba4d1228728e).
  - Refactoring sortable components/styles to be more generalisable between NameGeneratorPrompts/SociogramPrompts.
      - Sortable elements are now in their own sub-folder at: `src/components/StageEditor/Sortable/`
      - Interface remains the same:

      ```jsx
      // in NameGeneratorPrompts.js
      
      <FieldArray
        name="panels"
        component={Items}
        itemComponent={NameGeneratorPrompt}
      ...
      />
      ```
      - Added some new icons to UI for the styles

# Guidance component

We wanted to move to more flexible situation for interacting with help text (shown in the right hand column on Interface editors), where it could be triggered at the *section* level, but also for an individual field/component.

Decided on a redux driven <Guidance> component which appended the necessary events and referenced the content.

Redux was chosen so we weren't constrained in how this information was passed between edit components, an improvement might be to use React context, and limit the scope to the View which shows the side bar (and contains the editor).

A reference was chosen because a) it's not advised to store non-serialisable items in the store, and we needed formatting b) we know down the line we will be implementing locales, this seemed as good a time as any to start using them.

Guidance usage is like this:

```jsx
<Guidance contentId="my.item.from.the.copy.doc">
  Hovering this will set the guidance id to reference the content
</Guidance>
```

Locale usage looks like:
```jsx
// in locales/en_US.js
export default {
  'my.item.from.the.copy.doc': (
    <Fragment>
      <p>
        A good stage title should be descriptive, but not too long. It should help you to remember
        the purpose of this stage later.
      </p>
        ...
    </Fragment>
  ),
};
```

Current locale is stored in the state and currently hard-coded to enUS. A selector `getContent(state, contentId)`, allows us to fetch a content item by id, and filters to the respective locale data from the state.


src/components/Form/Fields/File.js
```diff
@@ -43,23 +45,33 @@ class FileInput extends PureComponent {
...
+        { label &&
+          <div className="form-fields-file__label">{label}</div>
+        }
+        <div className="form-fields-file__dropzone">
+          { value &&
+            <div className="form-fields-file__dropzone-preview">
+              {this.props.children(value)}
+            </div>
+          }
+          <Dropzone
```
bryfox: Probably not part of this PR, but when I drop new images on the sociogram background, the previous one isn't removed from storage. Might be nice to delete, or highlight that there's an unused image taking up space. Let me know if I should file an issue.

src/components/Form/Fields/Select.js
```diff
@@ -32,10 +40,18 @@ class Select extends PureComponent {
...
+      <div className={componentClasses}>
+        { label &&
+          <label
+            htmlFor={this.id}
+            className="form-fields-select__label"
+          >{label}</label>
+        }
+        <select className="form-fields-select__input" {...input} {...rest}>
+          {children}
+          {options.map(option => <option value={option}>{option}</option>)}
```
bryfox: I see a propType warning here when I first come to the screen: `Warning: Each child in an array or iterator should have a unique "key" prop.`

src/components/Guidance.js
```diff
@@ -0,0 +1,43 @@
...
+    setGuidance: PropTypes.func.isRequired,
+    clearGuidance: PropTypes.func.isRequired,
+    children: PropTypes.node.isRequired,
+  };
+
+  handleMouseEnter = () => {
+    this.props.setGuidance(this.props.contentId);
+  }
+
+  handleMouseLeave = () => {
```
bryfox: For form field-specific guidance, would this approach allow for focus/blur setting/clearing of guidance as well? (Seems like it...) Not needed in this PR, but would be nice to have for keyboard navigation of form.

bryfox:
One other comment: I did somehow get into a state where I had a final stage labeled "undefined interface" on the timeline, but I haven't been able to reproduce & can't give you any useful information about it. I don't have reason to think it's a regression.

src/styles/_settings.scss
```diff
@@ -1,5 +1,6 @@
 :root {
   --base-font-size: 15px;
+  --body-font-family:  -apple-system, BlinkMacSystemFont, 'Segoe UI', Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
```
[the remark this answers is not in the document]
wwqrd: I think Architect has slightly different font requirements? @jthrilly?

src/ducks/modules/root.js
```diff
@@ -24,6 +25,8 @@ const combineFilters = (...filters) =>
 export const rootReducer = combineReducers({
   form: formReducer,
   session,
+  locale: () => 'enUS',
```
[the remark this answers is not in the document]
wwqrd: 👍

src/utils/webShims/mockProtocol.json
```diff
@@ -658,9 +668,10 @@
       "type": "Information",
       "title": "Title for the screen",
       "items": [
-        { "type": "text", "content": "<p>Some arbitrary text content that can contain <strong>HTML</strong> tags.</p>" },
-        { "type": "audio", "content": "click_the_thing.mp3", "description": "Click by clicking the thing" },
-        { "type": "image", "content": "rubberduck.jpg" }
+        { "id": "1", "type": "text", "content": "<p>Some arbitrary text content that can contain <strong>HTML</strong> tags.</p>" },
```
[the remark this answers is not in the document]
wwqrd: At the moment I don't think these are used in NC at all, but they do make changes easier in Architect.

Not sure what the right call is here - I wonder if it's worth thinking about automatically adding and stripping these as required in Architect?

src/components/StageEditor/sections/SociogramPrompts/SociogramPrompt.js
```diff
@@ -0,0 +1,287 @@
...
+            <Field
+              name="text"
+              component={Fields.TextArea}
+              className="stage-editor-section-prompt__setting"
+              label="Text for prompt"
+              placeholder="Enter text for the prompt here"
+            />
+          </div>
+          <div className="stage-editor-section-prompt__group">
+            <h4 className="stage-editor-section-prompt__group-title">Nodes</h4>
```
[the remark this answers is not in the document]
wwqrd: I think this deserves it's own issue. The UI should really be using the actual nodes - and like you mention, is a required field. In NameGenerator we hide the rest of the form until NodeType is selected, perhaps we should do that here too.

https://github.com/codaco/Architect/issues/112

src/components/StageEditor/sections/SociogramPrompts/SociogramPrompt.js
```diff
@@ -0,0 +1,287 @@
...
+              parse={value => ({ type: value, entity: 'node' })}
+              format={value => get(value, 'type')}
+              options={nodeTypes}
+              label="Which node would you like to layout?"
+            />
+            <Field
+              name="sortOrderBy"
+              component={Fields.OrderBy}
+              className="stage-editor-section-prompt__setting"
+              variables={['*', ...variablesForNodeType]}
```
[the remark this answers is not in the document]
wwqrd: It's different in that it can be ascending/descending. Definitely confusing.

I will add some explanation to the guidance for now.

New issue: https://github.com/codaco/Architect/issues/113

src/components/Form/Fields/File.js
wwqrd: Good spot but we have an issue for this already!: https://github.com/codaco/Architect/issues/25

src/components/Form/Fields/Select.js
wwqrd: 👍

src/components/Guidance.js
wwqrd: That's a cool idea, I might try adding that functionality with a demo to the PR

wwqrd:
> One other comment: I did somehow get into a state where I had a final stage labeled "undefined interface" on the timeline, but I haven't been able to reproduce & can't give you any useful information about it. I don't have reason to think it's a regression.

I can't reproduce this either, my guess is that somehow the insert stage drawer returned undefined when creating a new one?

src/components/Guidance.js
wwqrd: So after a bit of experimentation with this I think it needs to include some kind of history.

So when you focus an element, then mouse over something else, when coming back to the element it can reset to the focussed guidance.

Will make this a separate issue.

PR #117, opened 2018-06-27
Shared <ControlBar /> component for use in <Card /> and <Protocol />

### LINKED ISSUE #85: Control bars are subtly different between Timeline and Stage edit
control bars

- Dimensions differ
- Color differs
- Padding differs
- The contained buttons are different sizes.
- Button margins differ

wwqrd:
Height difference still visible due to issue with <Button />, which
has a different height dependent on whether it contains an <Icon />.
This should be resolved in UI, will create a new issue for that.

Resolves #85


wwqrd:
Button height issue referenced here: https://github.com/codaco/Network-Canvas-UI/issues/64

PR #122, opened 2018-06-29
[WIP] Style changes for sunbelt

jthrilly:


wwqrd:
Looks good, do you have something in mind for the expandable variant of the sortable item too?

jthrilly:
> Looks good, do you have something in mind for the expandable variant of the sortable item too?

Yep, definitely. Just not got to it yet. Will homogenize with the other styles.

Do you have any thoughts on the modal for the variable chooser? Right now it seems unique to this instance. Did you have another modal system in place already that it should integrate with?

PR #125, opened 2018-07-06
[WIP] Feature/form editor reworked routes

wwqrd:
This adds a form editor screen, which can currently only be accessed from the timeline overview.

You should be able to create new forms, and edit existing ones. Those forms should then be visible for use in the name generator.

In future you should be able to add forms from the name generator (or other interfaces that use forms), but this isn't implemented yet.

### Known issues
- Correct input types for variable types
- Various UI issues from out-dated UI branch (missing icons, grey node text). These should be resolved when that branch is merged.
- Link up "create new form..." in NameGenerator using new routes
- Doesn't handle non-unique form names

Resolves https://github.com/codaco/Architect/issues/13


wwqrd:
This one was a bit chaotic due to the routes/view changes and could do with some tightening up RE tests etc during QA next week.

bryfox:
Looks good so far — transitions are seamless 👍 .

I focused on testing the new/edit form flow. Here are some UX things that I noticed; they may not necessarily be part of this PR (and sorry if you're aware of them all):

- Maybe hide the Form variables section until node type is chosen?
- I haven't been able to get my saved form to be available as a custom form on the name generator's stage editor. Should that be working yet?
- Forms not saved after clicking Save on form screen
  1. Create form (add title, type, and field)
  2. Click save
  3. On protocol screen, reload.
  4. Protocol is missing
  5. (It persists if, before step 3, I create/edit a stage so that I get the other Save button, and click that)
- Trying to save form without a title creates undefined errors
  - action: `{"type":"CREATE_FORM","form":{"optionToAddAnother":true,"fields":[{"variable":"age"},{"variable":"name"}],"type":"person"}}`
  - form errors on console: `[[undefined, undefined]]`
- Should the variable selects have empty options? Adding a field and leaving the defaults ("name", "TextInput") results in an empty field added (even though the UI indicates otherwise)
- I'm missing icons (or something?) for the handle, guidance toggle, and other controls... I just see the green background. UI submodule is updated.

missing-icons

wwqrd:
Thanks, I'll look into those!

> I haven't been able to get my saved form to be available as a custom form on the name generator's stage editor. Should that be working yet?

It should be available to select, but not editable yet. Maybe this is related to the saving issue.

> Should the variable selects have empty options? Adding a field and leaving the defaults ("name", "TextInput") results in an empty field added (even though the UI indicates otherwise)

They shouldn't have empty options, seems like a bug.

> I'm missing icons (or something?) for the handle, guidance toggle, and other controls... I just see the green background. UI submodule is updated.

This feature relies on a branch of UI which isn't merged into master, and that doesn't have the icons yet - so I expect that to resolve itself eventually.

bryfox:
> They shouldn't have empty options, seems like a bug.

They don't have empty options, so no bug there. The potential UX problem I was seeing was that (until you change the value) the selects display their first options, but the underlying data doesn't reflect that. e.g., you can save (and re-open) a form that looks like it has a "name" textInput, but doesn't actually define any fields.

name

```json
{"type":"CREATE_FORM","form":{"optionToAddAnother":false,"type":"person","fields":[{}],"title":"myform"}}
```

I was thinking that adding empty/placeholder/"Select one.." fields would be one way to solve the issue.

wwqrd:
@jthrilly I'd like to merge this in, uncompleted, with references to the the form editor removed as it contains quite a few changes which will be relevant to the variable registry editor.

PR #152, opened 2018-07-17
Refactor webpack config based on Server

bryfox:
- Update webpack plugins & loaders
- uninstall react-tap-event-plugin (unneeded)
- replace extract-text-plugin with minicssextract (for webpack 4)
- included "limit: 1" option for url-loader (for CSP)
- removed custom font alias for css-loader; seems unneeded
- added thread-loader plugin to parallelize babel
- moved browser config to package.json; target only Chrome version from Electron

You'll see a couple of peer dependency warnings after install that weren't there before, but (AFAICT from Server), they're nothing to worry about.


config/webpack.config.dev.js
```diff
@@ -82,206 +88,22 @@ module.exports = {
...
-
-      // First, run the linter.
-      // It's important to do this before Babel processes the JS.
-      // {
-      //   test: /\.(js|jsx)$/,
-      //   enforce: 'pre',
-      //   use: [
-      //     {
-      //       options: {
-      //         formatter: eslintFormatter,
```
bryfox: webpack.config.js re-enables the linter; if you want to comment it out again, see https://github.com/codaco/Architect/pull/152/files#diff-a58d55bdb5770c78ad512f8e91f8d051R105

wwqrd: I personally prefer having this stuff be part of my git push hooks - my editor shows the warnings either way. Otherwise I find it constantly breaking as a move things around. How do you feel about the lint settings?

bryfox: Yeah, I've configured my editor to insert "// eslint-disable-line" every time I type a newline.

Kidding aside, I love the change in `1a006e1` — I didn't know that was possible. I'm going to add that to Server, and would fully support the same in NC.

wwqrd: Where there's a will there's a way 😂

PR #158, opened 2018-07-30
Feature/form editor

wwqrd:
Looking for feedback on this, re transitions.

This feature re-implements links to the form editor (which was removed when the variable registry was being developed).

I've been experimenting with the animation between these cards and the protocol, but haven't found anything satisfactory. The zoom animation is particularly awkward to get right so I've held off on adding that to the protocol overview, which I'm under the impression is likely to change.

Things to look at RE transitions:
- Card transitions:
  **in**
  i.e. `Protocol -> Variable registry`, or `Variable registry -> Node type`
  Currently this is a simple fade in.
  **out**
  i.e. `Variable registry -> Protocol`, or `Node type -> Variable registry`
  This is a wipe out, probably we should mirror the stage transition and use a fade for "save"
- Zoom transitions:
  For `Variable registry -> Node type`, this is the growing blot of color when clicking a node type *icon*.
  I'm not sure this makes sense for clicking on the node type *name*, and for consistency I'm inclined to omit it from both *name* and *icon.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11899,6 +11899,24 @@
                 "prepend-http": "1.0.4",
                 "query-string": "4.3.4",
                 "sort-keys": "1.1.2"
+            },
+            "dependencies": {
+                "query-string": {
+                    "version": "4.3.4",
+                    "resolved": "https://registry.npmjs.org/query-string/-/query-string-4.3.4.tgz",
+                    "integrity": "sha1-u7aTucqRXCMlFbIosaArYJBD2+s=",
+                    "dev": true,
+                    "requires": {
+                        "object-assign": "4.1.1",
+                        "strict-uri-encode": "1.1.0"
+                    }
+                },
+                "strict-uri-encode": {
+                    "version": "1.1.0",
+                    "resolved": "https://registry.npmjs.org/strict-uri-encode/-/strict-uri-encode-1.1.0.tgz",
+                    "integrity": "sha1-J5siXfHVgrH1TmWt3UNS4Y+qBxM=",
+                    "dev": true
+                }
             }
         },
         "npm-run-path": {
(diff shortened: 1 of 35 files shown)


wwqrd:
Perhaps if the different "depth" screens had different colors, the transitions would be better pronounced?

bryfox:
The `Protocol -> Variable registry` fade looks nice. As an aside, once I have multiple node and edge icons in there, I wonder why they're all separate links (versus clicking anywhere in the rectangle), since they all link to the single registry. If the entire card (round rect) acted link a link, then zoom might look nice too.

I agree the out transition might be better as a fade; in general, I think symmetrical transitions make sense.

I think the `Variable registry -> Node type` feels a little jarring since it's a zoom followed by a fade; it's like it zooms in to nothing. (Maybe that was part of the difficulty?) But I think fade would be appropriate for this transition instead of zoom anyway; the node button doesn't really imply "container" (that I want to zoom in on) to me. I think that's what you're getting at with the *name* link, and I agree there.

wwqrd:
> As an aside, once I have multiple node and edge icons in there, I wonder why they're all separate links

They used to link directly to the individual types, this was just a quick implementation for the overview panel which is going to change. Until that happens, I agree that a single link zoom would look best.

> I think the Variable registry -> Node type feels a little jarring since it's a zoom followed by a fade; it's like it zooms in to nothing. (Maybe that was part of the difficulty?)

I'll remove it for now, it might be appropriate for other transitions (to the same screen), but it isn't in this case.

wwqrd:
- use form titles in overview panel
- if form title has space in it, it won't be saved to protocol automatically
- stage editor creates new stages on edit?

wwqrd:
- Add colors to protocol template

PR #165, opened 2018-08-03
Confirm and save before opening new protocol

wwqrd:
This is a somewhat ugly fix for the issue of confirming & saving before switching to a new protocol #65. At the moment, this seems like the most straight-forward fix.

It makes me wonder if literally using routes to manage views is the correct approach, or whether we should moved this into the app state. Perhaps even using a pseudo location, and pseudo <Route /> components - this would allow us to make app navigation happen more coherently in actions/thunks.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12450,21 +12450,10 @@
             "dev": true
         },
         "path-to-regexp": {
-            "version": "1.7.0",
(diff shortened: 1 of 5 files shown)


jthrilly:
I don't know...this looks good to me! Don't forget that the route driven stuff is good for animation and deep linking!

wwqrd:
I agree, we could emulate that -- effectively move the "store" from the history object, into redux. That'd mean we could tie it into our other actions more seamlessly.

PR #171, opened 2018-08-06
Set empty CSP values by default

bryfox:
In production, these should resolve to nothing; in dev, they're overridden by the build script. It adds the defaults using .env (to match Server). If you don't want to introduce that file, we can set the same values in the build script instead.

This isn't critical, it just prevents the console warnings in production.


wwqrd:
Unless it's something we might want to configure, shouldn't this go in the script?

bryfox:
It is configuration; the `env.raw` in `new InterpolateHtmlPlugin(env.raw)` is populated from those env files. I had to add the inlineCSP to the script since it wasn't static originally. (Though now, and if we're not able to get back to that dynamic approach, we could move it into env.development so the dev/prod scripts are the same).

I don't have a strong preference, though; I pushed the script version to `fix/csp-warning-inline` if you want to use it instead.

wwqrd:
Sorry, I don't mean it shouldn't be an ENV variable, just that it might not belong in `.env` - I assumed by build script you meant in `scripts/build.js`:

```
process.env.REACT_APP_SCRIPT_SRC_CSP = '';
```

On second thoughts though I think using `.env` is probably standard fare and less opaque.

PR #193, opened 2018-08-08
Update skip logic screen to use <Route />

### LINKED ISSUE #175: configure skip logic screen not opening
When clicking on "configure skip logic" for an interface, a yellow screen animates briefly to fill the screen, and then animates away again leaving you on the main protocol screen. I am unable to configure skip logic.

wwqrd:
Resolves #175

### DIFF
--- a/src/components/Cards/EditSkipLogic.js
+++ b/src/components/Cards/EditSkipLogic.js
@@ -2,15 +2,13 @@ import React, { PureComponent } from 'react';
 import { connect } from 'react-redux';
 import { bindActionCreators, compose } from 'redux';
 import PropTypes from 'prop-types';
-import { get, toPairs, find } from 'lodash';
+import { get, find } from 'lodash';
 import { Button } from '../../ui/components';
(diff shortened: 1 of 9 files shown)


wwqrd:
One of the rules from the development protocol is causing the editor to error:

Nothing has changed in the editor, so unsure what's causing this.

```
           {
              "type": "edge",
              "id": "15218241565624",
              "options": {
                "type": "friends",
                "operator": "EXISTS"
              }
            }
```

wwqrd:
Looks like it's missing an option, manually creating a similar rule in the editor also requires a variable to be set:
screen shot 2018-08-08 at 12 18 02

jthrilly:
> So the question is, which is correct?

We concluded that the dev protocol functionality (that is, optionally omitting a variable from an edge query) is "correct", but that the architect builder doesn't currently support it.

This will be added in #194, and the limitation is considered known for Alpha 3.

> Does the development example actually work?

Under investigation in NC repo.

PR #203, opened 2018-08-09
Fix/start screen animation

### LINKED ISSUE #200: Visual glitch on protocol stack transitioning from start screen to protocol screen
Overview card "blinks" before showing, instead of fading in.

wwqrd:
Revise screen animations to animate properly.

Long term, this is just too brittle to maintain. Any ideas for alternatives?

Required features:
- Animate enter/exit components from dom
- Tween nested components from one exiting component, to an entering component.

Resolves #200


wwqrd:
Animation appears janky when files are opened for the first time. Perhaps to do with file loading / extraction. I think this is passible, but like the above comments, it needs more thought.

Issue #205, opened 2018-08-10
white screen after changes


rebeccamadsen:
On windows, when I make changes to my protocol now (registry, forms, or interfaces), I get this error:

```
fs.js:646 Uncaught (in promise) Error: ENOENT: no such file or directory, open 'C:\protocol.json'
    at Object.fs.openSync (fs.js:646:18)
    at Object.module.(anonymous function) [as openSync] (ELECTRON_ASAR.js:172:20)
    at Object.fs.readFileSync (fs.js:551:33)
    at Object.fs.readFileSync (ELECTRON_ASAR.js:510:29)
    at loadProtocolData (http://localhost:3003/static/js/bundle.js:152976:63)
    at http://localhost:3003/static/js/bundle.js:149926:97
    at http://localhost:3003/static/js/bundle.js:108677:18
    at dispatch (http://localhost:3003/static/js/bundle.js:109094:18)
    at http://localhost:3003/static/js/bundle.js:149941:7
    at <anonymous>
fs.openSync @ fs.js:646
module.(anonymous function) @ ELECTRON_ASAR.js:172
fs.readFileSync @ fs.js:551
fs.readFileSync @ ELECTRON_ASAR.js:510
loadProtocolData @ loadProtocolData.js:10
(anonymous) @ file.js:74
(anonymous) @ index.js:8
dispatch @ applyMiddleware.js:35
(anonymous) @ file.js:90
Promise resolved (async)
(anonymous) @ file.js:87
(anonymous) @ index.js:8
(anonymous) @ bindActionCreators.js:3
(anonymous) @ ProtocolLoader.js:31
callCallback @ react-dom.development.js:10878
commitUpdateQueue @ react-dom.development.js:10911
commitLifeCycles @ react-dom.development.js:14377
commitAllLifeCycles @ react-dom.development.js:15462
callCallback @ react-dom.development.js:100
invokeGuardedCallbackDev @ react-dom.development.js:138
invokeGuardedCallback @ react-dom.development.js:187
commitRoot @ react-dom.development.js:15603
completeRoot @ react-dom.development.js:16618
performWorkOnRoot @ react-dom.development.js:16563
performWork @ react-dom.development.js:16482
performSyncWork @ react-dom.development.js:16454
requestWork @ react-dom.development.js:16354
scheduleWork$1 @ react-dom.development.js:16218
enqueueSetState @ react-dom.development.js:11299
./node_modules/react/cjs/react.development.js.Component.setState @ react.development.js:270
[...]

PR #206, opened 2018-08-13
Fix/alpha 3.1

jthrilly:
There were some pretty bad bugs in this release that I introduced with too many last minute changes, which needed to be fixed before a demo planned for this Wednesday. I am leaving the fixes in a PR, because I suspect there will be a better way to solve some of these issues.


jthrilly:
> I'm not sure about better ways to solve issues, but I can say that this solves #205 for me.

Thanks for checking this out for me, Rebecca! Glad it has worked.

jthrilly:
Discussion with Steve regarding this:

- layout -> layoutVariable is an undesired change, and in fact we were looking at reversing this. NC needs an issue creating to do this.
- White screen after closing the edit screen will be resolved by changes to timeline animation.
- Displaying appropriate node labels and colors is a valid change, and will be resubmitted in a new PR.
- Other small content edits will also be submitted in new PRs.

Issue #212, opened 2018-08-23
Review error handling


wwqrd:
Meta-issue for:

https://github.com/codaco/Architect/issues/133 - Errors throughout SociogramPrompt need some thought
https://github.com/codaco/Architect/issues/169 - Errors are shown right away on new/edit forms
https://github.com/codaco/Architect/issues/135 - Sociogram is missing validations
https://github.com/codaco/Architect/issues/134 - Name Generator is missing validations

wwqrd:
Partially resolved by https://github.com/codaco/Architect/pull/213

Outstanding issues:
- Force errors to show on loading an invalid stage? (rather than waiting for touched signal)
- Stylistic options missing from required fields
- Animation to highlight element on scroll to?

wwqrd:
Resolved by #215

PR #213, opened 2018-08-28
[WIP] Feature/review error handling

wwqrd:
This is a work in progress to tackle the two issues in #212.

Only tackles stage editor at the moment, but could be expanded to other sections.

Changes:
- On stage editor, submit button is not 'disabled' for an invalid state. Instead, once a user clicks continue, they are shown an interactive panel listing the errors on the page.
- Issues may be clicked to go to the respective error.
- Expandable items are forced open if they contain errors - not totally sure about this interaction.
- Issues panel can be collapsed to the title.

Known issues:
- field names aren't friendly, maybe don't show them?
- If a field is loaded that contains errors, it can mess up the field rules for showing errors:
  `touched && errors`


bryfox:
I like this — seems like a good improvement for usability. 

- I agree that field names aren't friendly, but it seems like something needs to display there. Is there an easy way to get at the label text? Or is it feasible to rely on convention and map simple patterns like `prompts[0].displayName` -> `Prompt 1 display name`?
- Looks like "required" is no longer red; was that intentional? Color (or some similar styling) might help call out the problem better in addition to the scrolling.
- Is it worth adding icons to make the issues & header more obviously clickable? e.g., a minimize icon in the upper right of the header bar?

wwqrd:
Thanks @bryfox, I've implemented some changes based on that feedback:
- Labels, since the title is contained in the presentation components (DOM), have used refs and post render updates to fetch label text.
- Have added error styling to the markdown field type
- Added some visual flourishes to the issue title bar, including showing the open/close state.

screen shot 2018-08-29 at 14 33 55

src/components/Issues.js
```diff
@@ -0,0 +1,161 @@
...
+        return (
+          <li
+            key={fieldId}
+            className="issues__issue"
+          >
+            <a
+              href={`#${fieldId}`}
+              onClick={this.handleClickIssue}
+            >
+              <span ref={el => this.setIssueRef(el, fieldId)}>
```
bryfox: This span changes the event target ([L91](https://github.com/codaco/Architect/pull/213/files#diff-ec976aaee2548bdfb5fb02cc65e88f5dR91)); scrolling fails when I click on the text itself.

bryfox:
I added an inline comment about the added `span`, but otherwise this looks great to me.

PR #214, opened 2018-09-04
[WIP] Fix/variable renaming

### LINKED ISSUE #190: Variable chooser is bugged
The variable chooser for prompt variables seems to not behave correctly. 

- Boolean variables show '0' and '1' as potential values, rather than true/false.
- The input label is white, which looks bad on the light gray background.
- The button label should only change to "Done" when a value has been selected/changed. Perhaps simpler to just show a cancel button at all times, which would remove the need for this logic.

### LINKED ISSUE #208: Editing a variable after it has been added to a form can create invalid states
If I edit the nickname variable (say by renaming it to "display name"), it seems to update correctly in the "add a new person" form, but in Network Canvas the form displays without a label. 

Removing the variable from the form and adding it again corrects this issue.

wwqrd:
Fixes #208 by adding variable name as a property in the registry, instead of using the key as both an id and a name.

IDs/keys are now randomly generated, and due to the way Network Canvas works this means that these keys will be used when assigning properties to node in the network, until it is updated in https://github.com/codaco/Network-Canvas/issues/648:

```
// in variableRegistry
"node": {
  "person": {
    "variables": {
      "45745c60-7b1a-11e8-9c9c-2d42b21b1a3e": {
        "name": "nickname,
        ...
      }
    ...
  }
  ...
}

// in Network Canvas
network: {
  nodes: [
    { type: 'person', '45745c60-7b1a-11e8-9c9c-2d42b21b1a3e': 'Bobby' },
    ...
  ],
  ...
},
```

Also resolves #190


rebeccamadsen:
I think this change makes sense, and I like the attribute table.

I'm not sure I'm testing everything, but when I try to select an "additional attribute", I get this error and nothing selects:
```
Uncaught TypeError: Cannot convert undefined or null to object
    at hasOwnProperty (<anonymous>)
    at main.09bf9118.js:148
    at Object.createVariable (main.09bf9118.js:125)
    at n.handleChooseVariable (main.09bf9118.js:148)
    at onClick (main.09bf9118.js:148)
    at l (main.09bf9118.js:38)
    at Object.<anonymous> (main.09bf9118.js:62)
    at Object.invokeGuardedCallback (main.09bf9118.js:62)
    at Object.invokeGuardedCallbackAndCatchFirstError (main.09bf9118.js:62)
    at C (main.09bf9118.js:62)
```

I tried resetting app data, and creating a new protocol.

Also, the names for the Person node start out empty, so it isn't valid on load. You have to go through and name all the variables just to change the node color.

wwqrd:
We're going to merge this this release, and handle the Network Canvas side (transposing on export: https://github.com/codaco/Network-Canvas/issues/648) next release.

PR #219, opened 2018-09-06
New form selector

### LINKED ISSUE #136: "No form" should be an option in Name Generator
Some NameGenerators will be entirely roster driven (e.g. from external data or previously defined nodes) and won't need to create nodes using a form.

This will probably require "a form or at least one panel" for validation.

Could pair well with integrating the form editing feature with Name Generator.

wwqrd:
Implements the following interface for forms:

screen shot 2018-09-06 at 16 47 06

Known issues:
- Doesn't implement "default" form.
  - We no longer have the concept of a default form in Network Canvas since we can now have "no form", i.e. null/undefined as a form option.
  - Selecting a nodeType will affect the choice of default, making it awkward to implement in redux-form
  - Instead, this has been made a required field, so the user must explicitly choose from either an existing form or no form.

Resolves #136


wwqrd:
On further inspection I see what you mean. If the `form` prop isn't defined, then we default to nodeType, if `form: null` - then we don't display a form.

I'm not sure what the right approach is here, but I think that's a not a great interface. I'd be tempted to either: a) require the prop to be defined and have a special value for 'default', b) or drop the 'default' feature, since it's so simple to require the field in Architect.

@jthrilly?

jthrilly:
> drop the 'default' feature, since it's so simple to require the field in Architect.

I think what you have implemented here in Architect basically _is_ the default feature (which is kind of your point). So we can remove the logic for this from NC.

wwqrd:
I'll merge this in, since it should still work as expected in NC with the prop always being defined - and add a comment to https://github.com/codaco/Network-Canvas/issues/592

PR #235, opened 2018-09-12
don't check for updates on startup

### LINKED ISSUE #236: "Architect is up to date" shouldn't be displayed on startup
App checks for updates on startup - this should only alert the user when an update is found.

If the user checks manually - then the "up to date" dialog is appropriate.

jthrilly:
Fixes #236


jthrilly:
Fixes #236

PR #242, opened 2018-09-20
Feature/delete type side effects

### LINKED ISSUE #232: In use edges/nodes can be deleted, breaking related interfaces
It is possible to delete edges and nodes that are in use, which produces invalid interfaces (https://github.com/codaco/Network-Canvas/issues/673, https://github.com/codaco/Network-Canvas/issues/674).

Some possible solutions:
- Allow deletion, but mark protocol as invalid and prevent saving until it's fixed.
- Prevent deletion until related interfaces are changed.

wwqrd:
When deleting a type, checks for usage and deletes any affected objects.

Implemented as a switch on the deleteType action `deleteType(entity, type, deleteRelatedObjects = false)`

- [x] When deleting a type notify user of the impact on other objects.
  - Currently implemented as a confirm dialog - thinking this might be better as an in-app modal so that we can do more with UI.
- [x] Delete any associated stages/forms.
- [x] Add information to node/edges registry indicating where assets are used

Resolves #232

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
         "dist:win": "npm run build && electron-builder -w --publish=always",
         "dist:all": "npm run build && electron-builder -mlw --publish=always",
         "bump": "sh scripts/bump",
-        "preflight": "[[ -z $(git status --porcelain) ]] && npm run lint && npm run lint:sass -- --max-warnings 0 && npm test -- --coverage"
+        "preflight": "npm run lint && npm run lint:sass -- --max-warnings 0 && npm test -- --coverage"
     },
     "repository": {
         "type": "git"
--- a/src/components/Cards/VariableRegistry.js
+++ b/src/components/Cards/VariableRegistry.js
@@ -72,16 +72,15 @@ class VariableRegistry extends Component {
       '',
     );
 
-    if (usage.length > 0) {
-      // eslint-disable-next-line no-alert
-      if (!confirm(`
-Because a number of other objects depend on this type, they will also be removed:
-${deletedObjects}
-      `)) { return; }
-    }
+    const confirmMessage = `Are you sure you want to delete "${type} ${entity}" ${
+      usage.length > 0 ?
+        `\n\nBecause a number of other objects depend on this type, they will also be removed: \n${deletedObjects}` :
+        ''
+    }`;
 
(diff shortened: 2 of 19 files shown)


wwqrd:
This only flags unused types, since we have another issue already that includes unused variables, and the functionality is complex enough to warrant it's own PR.

wwqrd:
RE in-app modals vs confirm() dialog, we also have a separate issue fo this, so will implement this as a separate PR too.

rebeccamadsen:
I think this works really well. I feel like the error messages could be more informative, though that could be a follow up PR as part of moving to an in app modal. For example, I wasn't clear when a prompt told me a stage prompt would be removed that the entire stage would be (because it was the only prompt). It was what I expected to happen as a developer, but the message itself didn't indicate that it would.

If they were richer modals, it might be nice to allow clicking through to the prompt or form in question?

Deleting a type that led to a form being removed gave me the form name, but a type that led to the prompt being removed gave me two uid's instead, which isn't as helpful:

screen cap

I think we don't need two confirmation dialogs; if I approve the stage/form being deleted, can that be enough to just go ahead and delete the type as well?

wwqrd:
> Deleting a type that led to a form being removed gave me the form name, but a type that led to the prompt being removed gave me two uid's instead, which isn't as helpful:

Agreed, I'm going to update the mockProtocol with uuids to make this more obvious if it happens in future, and make the notice use the labels 👍

> For example, I wasn't clear when a prompt told me a stage prompt would be removed that the entire stage would be (because it was the only prompt). It was what I expected to happen as a developer, but the message itself didn't indicate that it would.

You're right the user really needs to know that upfront! I'll add a check and maybe update the deleteType action to skip the prompt delete step where possible.

> I think we don't need two confirmation dialogs; if I approve the stage/form being deleted, can that be enough to just go ahead and delete the type as well?

I guess it is a little over cautious! We'll have undo functionality soon, so this probably could go.

rebeccamadsen:
This is much more intuitive! I think if the prompt delete message could also indicate which stage that prompt is on, it would be easier to find. Although it's possible that's a developer-centric issue since I'm using gibberish in my names, so find it harder to remember which prompt goes with which stage!

PR #244, opened 2018-09-27
Fix logger setup

### LINKED ISSUE #243: App errors on very first open
The app's data directory isn't created until runtime, and the logger tries to access a file in the directory before it exists. This results in an error alert for a first-time user. (Tested macOS.)

```
Uncaught Exception:
Error: ENOENT: no such file or directory, open '[...]/Library/Application Support/Network Canvas Architect/debug.log'
```

Steps:
1. Uninstall Architect (really, just remove its data directory)
2. Reinstall
3. Open the app

bryfox:
Fixes #243.

This uses the [default log location from electron-log](https://github.com/megahertz/electron-log/blob/fc1c768/lib/transports/file/find-log-path.js#L15-L54), which is the most straightforward fix.

If the log needs to be written into the userData dir, then we'll need to either:
- rely on electron's `getPath` but ensure that log setup (and, consequently, logging) calls are only made after electron's ready event, or
- implement our own directory resolution like electron-log's above, and create that location if it doesn't exist (knowing that electron will also be looking to create it)

Other minor changes:

- Mode changed from 'w+' to 'w'; logger shouldn't need read access. (Note: the default behavior would be to append.)
- Update require in electron-starter for clarity; custom setup needs to happen before the first logging call.

### DIFF
--- a/public/components/log.js
+++ b/public/components/log.js
@@ -1,15 +1,9 @@
-const electron = require('electron');
-const path = require('path');
-const fs = require('fs');
 const log = require('electron-log');
-
-const userDataPath = (electron.app || electron.remote.app).getPath('userData');
 
 log.transports.console.level = false;
 log.transports.file.level = 'log';
 log.transports.file.format = '{h}:{i}:{s}:{ms} {text}';
 log.transports.file.maxSize = 5 * 1024 * 1024;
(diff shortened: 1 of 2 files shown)


rebeccamadsen:
This works for me on Windows.

PR #248, opened 2018-10-08
Fix protocol creation

bryfox:
Fixes protocol creation errors, which were introduced with recent dependency changes. (The problem can be traced to webpack 4.20; however, this also downgrades Electron & Node for consistency with other apps.)

The behavior on master is a LOAD_CREATE_ERROR during archiving, which is triggered from a type check inside compress-commons (the Stream is no longer recognized as such.)

Steps to test (or see https://github.com/codaco/integration-tests/pull/3):
1. Install dependencies
2. Start app
3. Click "Create Protocol"


wwqrd:
👍

Issue #250, opened 2018-10-09
Filter UI displays ID instead of variable name


bryfox:
When adding a filter to a panel, the variable ID is displayed instead of its name.

![panel-ids](https://user-images.githubusercontent.com/39674/46682230-f637ae80-cbba-11e8-99c3-a2c8dde30cf7.png)

Steps taken:
- Create protocol
- Add NG interface
- Choose default Person type
- Add a Panel
- Add a filter

PR #257, opened 2018-10-10
Use variable names in filter

wwqrd:
Fixes: #250


wwqrd:
Ahh I see, updated for variable names, but not entity names, though there does seem to be another problem with that which I might address here, which is that entities are missing a name configuration option in the UI: https://github.com/codaco/Architect/issues/259

jthrilly:
@wwqrd good to merge?

PR #265, opened 2018-10-12
Fix form creation

### LINKED ISSUE #263: Cannot create new form
Steps:
1. Create a protocol
2. Click Create new form

Observed: cannot click a node type, so form cannot be created.

The Node component in ui no longer receives spread properties; the editor was relying on this for the click handler.

bryfox:
Fixes #263.

Captures the click above the UI component, since `<Node />` no longer receives arbitrary props.

Regression test added to [integration-tests](https://github.com/codaco/integration-tests/pull/4).


rebeccamadsen:
Looks good to me!

PR #278, opened 2018-10-15
Show entity names in variable type delete dialog

### LINKED ISSUE #269: delete edge and node use id
Probably they should use name instead of id since name is user facing elsewhere.

delete edge

wwqrd:
Fixes #269


rebeccamadsen:
Works for me!

Issue #275, opened 2018-10-15
Additional attribute removed if value is empty


bryfox:
Minor UX inconvenience: On a name generator prompt, while creating or editing an additional attribute, if the input value changes to an empty string, then the attribute is removed from the view entirely. While editing, I would not expect the form field to disappear. Example: create an attribute, type a letter, hit backspace.

Issue #264, opened 2018-10-12
Boolean additionalAttributes should default to true


bryfox:
When I add a boolean additionalAttribute (like "close friend"), it has an undefined value. In order to have it save as `false`, I need to toggle it on then off again, but there's no indication that this is required, or that the initial undefined state is different than `false`.

jthrilly:
I've noticed this, too. However I think it should default to `true` given the context.

jthrilly:
@wwqrd - just wanted to confirm that this was fixed in #284 (please close if it was).

wwqrd:
Yep this is fixed, boolean attributes default to true.

PR #284, opened 2018-10-19
Manually manage state in Attributes Table

wwqrd:
Due to the way redux-form manages it's state, empty fields in AttributesTable were being unmounted and lost. I think this comes down to an issue of using redux-form to drive dynamic forms, where the <Field /> components are actually generated reciprocally from the state.

This PR manages the AttributesTable value in component state, before calling redux-form's change()
method to update. It also introduces a DetachedField component which mirrors the one provided by
Redux Form, but doesn't connect to the redux state.

An alternative implementation might be to manage the variable references in component state, and
keep the values in redux-form, but that seems equally complicated if not more so.

Fixes: #275, #264


rebeccamadsen:
This helps with not removing an attribute. It makes me wonder what other forms we're using are funky in this way...but I guess in NC we don't have dynamic forms.

Should the validation rules for additional attributes be satisfied before you can save? I noticed that some additional attributes, the value was "required", but I could save anyway. Or on age, the value should be less than 100, but I could put 222.

Also, should boolean values default to true? I'm seeing false here.

wwqrd:
Ugh, not happy with this fix, but couldn't find a way around it.

We leverage Redux Forms for error handling on submit, so I've somewhat reinstated using additionalAttributes using Field (We used to use it for every input, now the entire table is treated as a single field). Field level validations are handled normally, with the emulated Field component (DetachedField), and form level validation is handled by performing one big validation on attributes table.

wwqrd:
Known issues:
- [ ] When field is closed for editing, can't see validation errors
- [ ] When multiple errors are contained in the attributes table, only one is shown in the issues panel

rebeccamadsen:
Looks good to me!

PR #285, opened 2018-10-19
Show Items errors on submit

### LINKED ISSUE #230: Error should show for field arrays on submit
Click submit on name generator with no prompts - error displayed in issues panel but not inline.

error not shown on submit for field array

wwqrd:
Fixes #230


rebeccamadsen:
Looks good!

PR #286, opened 2018-10-19
Replace new form plus button with text button

### LINKED ISSUE #229: Form 'plus' button, doesn't match meaning of FieldArrays
Suggest replacing with text "Create new form" button.

plus button doesn t make sense in contenxt of othre plus buttons

wwqrd:
Fixes #229


rebeccamadsen:
Looks good!

PR #291, opened 2018-10-25
Stage Previews

wwqrd:


wwqrd:
Remaining tasks:
- [ ] Move preview button to control bar
- [ ] Clean up preview window, remove green title bar for now, planned specific chrome
- [ ] Only allow one preview window to be open at a time. When navigating away, keep open until manually closed; when user clicks to preview another stage, update any open preview window first.

Known issues:
- SVG assets in Network Canvas break stylesheet build, currently using a branch of NC to build.

wwqrd:
Tested with 600 stage protocol, and performance seems snappy - is this enough of a load test? One thing I did notice about that is the timeline view was very slow, that could maybe do with some virtualisation (not rendering off-screen elements).

wwqrd:
New bug found, for new protocols, there is no stage id which is currently a dependency of the preview functionality - will look into alternative way of handling this.

wwqrd:
Review issues:
 
- [ ] Script submodule checkout (nested dependencies)
- [ ] Window brought to top when opening preview, but not forced on top.
- [ ] Put refresh/close buttons in control bar
- [ ] Node form won't open
- [ ] Navigating away from stage editing should close preview
- [ ] Can't preview new protocol, with new stage.
- [ ] Toggle control for different aspect size presets, and scaling.

PR #300, opened 2018-10-29
Fix name of Text component

bryfox:
Updates the Text component to match the [implementation in Network Canvas](https://github.com/codaco/Network-Canvas/blob/master/src/containers/Field.js#L27).


wwqrd:
I think this might be superseded by https://github.com/codaco/Architect/pull/303

PR #310, opened 2018-11-09
Add sorting / searching to variables list in the TypeEditor

wwqrd:
Adds new UnorderedList component, which can show a filtered
list of items through a FieldArray via an adapter.

Also renames Items -> OrderedList.

Resolves: https://github.com/codaco/Architect/issues/301


wwqrd:
- [ ] Use Text UI components
- [ ] Odds/evens results on search filter?
- [ ] Sort by match?

PR #341, opened 2018-11-29
[WIP] Make filter group more reuseable

bryfox:
https://github.com/codaco/Server/pull/197 is making use of this version of the components.

In short, this removes the external selector/redux dependency from the FilterGroup component tree. `variableRegistry` is passed in to FilterGroup as a component prop. This happens in two places:

- NodePanels -> NodePanel
- EditSkipLogic -> SkipLogicEditor

Other minor changes:
- make the close buttons actual `button`s so that they're keyboard-accessible
- remove a 'getAbsoluteBoundingRect' import which wasn't needed; this animation can be more easily shared in UI in the future.


wwqrd:
Resolved by #346

PR #342, opened 2018-11-30
Only show preview button when protocol is valid

### LINKED ISSUE #333: Error checking before preview?
On the stage editor, should the Preview button use the same error guard that Continue does (to prevent stage loading errors in the preview)?

wwqrd:
Also animates appearance of buttons.

Resolves  #333


wwqrd:
I think you're right. There's too much going on with this, and a disabled state would be less mental overhead.

PR #346, opened 2018-12-14
New filter/query UI

wwqrd:
This revised query UI uses text descriptions for the rule preview, and rule editing is done in a modal window.

Testing:
Best place to inspect it is when editing Name Generator panels or Stage skip logic.


wwqrd:
> - I'm not sold on the 'alter rule type' and 'alter variable type' distinction. Can we merge them into one dialog?

Do you have an idea of how you would like that to work UI wise?

> - Can we hide the rest of the form until the entity type is selected?

:+1:

> - The cancel and save buttons are the opposite way round to the cancel and preview buttons in the interface editor. One of them should probably change.

👍 

> - Exists/not exists is not quite the right way of describing the check on the attribute. For example, I chose "close friend", and it wasn't clear to me if I was testing the truthiness of the value. This also doesn't work for fields like "nickname".

This is a check for a variable being set, not truthiness. Now that we initialise variables when a node is created, we can remove this option.

> - I can't create a query about the value of a categorical or ordinal variable. I should be able to test the value of these, I think.

:+1:

jthrilly:
> Do you have an idea of how you would like that to work UI wise?

I might be missing something obvious, but they basically do the same "thing", except that they have different options in the first dropdown. These should be merged, and then the behaviour of the second dropdown should change according to what was selected first.

So for example if you choose an "exists" in the first dropdown, there's no need for a second one. 

If you choose a variable, you then get the second dropdown with the comparators.

wwqrd:
Need to discuss EXISTS/NOT_EXISTS, regarding nodes and edges too. e.g.

"Node type 'person' exists" doesn't seem the right way to describe "Is a 'person' node".

wwqrd:
> Need to discuss EXISTS/NOT_EXISTS, regarding nodes and edges too. e.g.
> "Node type 'person' exists" doesn't seem the right way to describe "Is a 'person' node".

After a call on this, we decided that it makes logical sense as is, just is unusual syntax from a human language perspective. Therefore we leave it as is and provide good documentation.

PR #353, opened 2018-12-18
Fixes issues with the new Roster List editors

### LINKED ISSUE #351: Bugs with the new roster edit interfaces
- [x] Name Gen List: The global boolean `showExistingNodes` should be a configurable property, defaulting to true.
- [x] Name Gen Autocomplete: Can't select display label for HIVServices external dataset.
- [x] Both: Preview mode doesn't work (looks to be an issue with fetching external data).

wwqrd:
- Added showExistingNodes option to the Name Gen List editor
- Display label, and other attributes updated to get properties from the variable registry,
rather than the dataset itself.
- Preview updated to support copying external data from protocol.

Known issues:
- Preview functionally works, but styles are slightly distorted. Suggest that this is fixed separately.

Resolves #351


wwqrd:
Thanks Bryan, I'll add some description text to the field.

wwqrd:
Found a bug with this related to externalData not having uids, will fix before merge.

wwqrd:
Okay, previews and externalData fixed, this should be ready to go.

rebeccamadsen:
Looks good to me.

PR #356, opened 2019-01-03
Categorical Bin editor

### LINKED ISSUE #16: Categorical Bin configuration screen

wwqrd:
Resolves #16

### DIFF
--- a/src/components/StageEditor/sections/CategoricalBinPrompts/CategoricalBinPrompts.js
+++ b/src/components/StageEditor/sections/CategoricalBinPrompts/CategoricalBinPrompts.js
@@ -1,0 +1,22 @@
+import React from 'react';
+import Prompts from '../../../Prompts';
+import CategoricalBinPrompt from './CategoricalBinPrompt';
+
+const CategoricalBinPrompts = props => (
+  <Prompts
+    contentId="guidance.editor.name_generator_list_prompts"
+    promptComponent={CategoricalBinPrompt}
+    template={{ showExistingNodes: true }}
+    {...props}
+  >
+    <h2>Prompts</h2>
+    <p>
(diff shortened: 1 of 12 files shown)


src/components/StageEditor/sections/CategoricalBinPrompts/CategoricalBinPrompts.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Prompts from '../../../Prompts';
+import CategoricalBinPrompt from './CategoricalBinPrompt';
+
+const CategoricalBinPrompts = props => (
+  <Prompts
+    contentId="guidance.editor.name_generator_list_prompts"
+    promptComponent={CategoricalBinPrompt}
+    template={{ showExistingNodes: true }}
```
bryfox: I don't see `showExistingNodes` supported in NC. Do we need a ticket to implement there?

wwqrd: That's a mistake, it's now been removed

jthrilly:
In ordinal bin we are missing the `color` attribute from each prompt, which is designed to allow the participant to create visual associations between certain types of questions and colors.

Can we add a color picker to the prompt, build on `ord-color-seq` and defaulting to `ord-color-seq-1` (without automatically incrementing.

jthrilly:
While we are looking at this, another small improvement to this particular prompt pattern would be to have a new prompt open in the expanded state. Is that possible?

wwqrd:
@jthrilly  The ordinal bin now includes the color picker, additionally created our own color picker component since react-color required a few work arounds to use with our variable based palettes.

PR #368, opened 2019-01-17
implement quick add property in name gen editor

jthrilly:
This PR implements the `quickAdd` property for the name generator interface, and updates the NC and UI dependencies.


config/webpack.config.js
```diff
@@ -255,7 +255,9 @@ module.exports = {
       // It is guaranteed to exist because we tweak it in `env.js`
       process.env.NODE_PATH.split(path.delimiter).filter(Boolean),
     ),
-    alias: {},
+    alias: {
+      concaveman: path.join(paths.appSrc, 'utils', 'webShims', 'concavemock.js'),
```
jthrilly: Incase you are wondering, this is needed because this module does an unsave-eval. 

Until we implement the narrative interface editor and therefore need to preview it, this is a workaround.

We also intend to v2.0 the preview function which will resolve this the proper way.

wwqrd:
Looks fine to me, do you think the quick add button should hide until a node type is selected or not?

jthrilly:
> Looks fine to me, do you think the quick add button should hide until a node type is selected or not?

Yes, I think so. If it doesn't at present, could you please add that?

PR #375, opened 2019-01-23
Improved prompt editing

wwqrd:
- Prompts are opened in a modal for editing, which makes the list view less cluttered
- When new prompts are added the edit window is opened automatically

Testing:
- Requires updated version of react-flip-toolkit, so run `npm install`

Known issues:
- [x] Prompt can't be edited if preview is blank.
- [x] Seem to be able to save stage when prompt has errors?
- [x] Cannot view sidebar information
  - Possible fix: add a portal to the main view in Card and allow items to render there, like the main stage editor
- [x] If you add a new prompt, then saving it fails, it remains (invalid) in prompts list.
- [x] Issue panel obscures form when open
- [x] Prompt submit buttons look terrible, perhaps they should exist in the bottom toolbar?
- [x] Errors in prompt are incorrectly styled
- [x] Form should probably have a title. Do we need to think about breadcrumbs in the UI?
- Other editor windows are now broken and need updating to new format:
  - [x] Entity editor
  - [x] Form editor
  - [x] Skip logic

Further possible improvements:
- [x] Don't allow modal to be closed if prompt contains errors/highlight prompts with errors
- [ ] Clearer scoping of props, we need the spread operator to make PromptForm extendable, but we should be clear about what else is passed along.
- [ ] Inconsistent validation warnings for prop types, should it be object or array? (May be unrelated to this PR, see react warnings)


wwqrd:
Because prompt fields are only shown in modal form wasn't validating these fields! Have implemented them as a sub-form which resolves: "Seem to be able to save stage when prompt has errors?" and "Don't allow modal to be closed if prompt contains errors/highlight prompts with errors"

wwqrd:
https://github.com/codaco/Architect/issues/370 May relate to the issue above with errors displaying incorrectly. Perhaps this should be handled as a separate PR against 370

jthrilly:
Big improvement! Particularly for the sociogram, this makes a ton more sense, I think.

The main issue I see is that I think we have a bit of confusion about if this is a separate screen or if its a modal overlay. At the minute it is halfway between. It seems to open "full screen" (implying it is a separate screen), but then it has its own cancel and continue buttons in addition to the stage level purple bar at the bottom (implying it is an overlay). A couple of times I instinctively clicked the cancel button in the bottom bar, and went out of the stage I was editing.

I think we should resolve this by making this more firmly a modal overlay. So have the panel expand only to a certain width and an auto height (with overflow scrolling), perhaps dim the rest of the stage, and keep the panel background color and animation from the summary state to emphasise the origin.

Small changes:
- rename "continue" -> save and continue
- textarea styling (rounded corners and padding to match other inputs. maybe add this as an input type to UI?

wwqrd:
I've updated this based on our catchup so that:
- Prompt appears inset in the left panel so that it appears more like a modal
- Control buttons are anchored at bottom
- If there is a validation issue, it will auto-scroll to the first issue.

screenshot 2019-01-28 at 19 41 42

wwqrd:
> - textarea styling (rounded corners and padding to match other inputs. maybe add this as an input type to UI?

This does look totally messed up at the moment. Given how huge this PR already is, I think it'd be better to handle that separately, along with the error styling. As far as I know it's not a side-effect of any of the changes here.

wwqrd:
UI update for text areas can be found here: https://github.com/codaco/Network-Canvas-UI/pull/96

PR #384, opened 2019-02-07
remove template protocol entities

jthrilly:
This PR removes the remaining data from the template protocol that new protocols are based on.


wwqrd:
Looks good.

PR #395, opened 2019-02-14
New Asset Browser component

### LINKED ISSUE #365: Create an inline asset browser
The upload field should be updated to contain a 'browse' link, which will let the user choose from existing files in the assetManifest:
screenshot 2019-01-08 at 11 49 29

Clicking browse should open a modal with a list of suitable assets from the assetManifest. Selecting one will close the window and set it as the asset for this field.

Part of #290

wwqrd:
Resolves #365


wwqrd:
Feedback from demo w/ @jthrilly :

- Radio group for Datasource w/ use network from interview
- Asset browse cancel button right
- No blur on hover text (or maybe add icons to clarify)
- Only panels allow selecting interview network.

PR #409, opened 2019-03-08
Support for editing Narrative Interface

wwqrd:
New narrative interface editing capability.

## Known issues
- [ ] In presets, what should the spec be for empty values, omission? Currently values are templated as:

  ```
  const narrativePresetTemplate = {
    layoutVariable: null,
    groupVariable: null, // When left empty, this is set to "" - do we need to clear it?
    edges: {
      display: [],
    },
    highlight: [],
  };
  ```
- [ ] Guidance text needs to be set for new interface sections.

Resolves: https://github.com/codaco/Architect/issues/287

### DIFF
--- a/src/components/EditableList/withEditHandlers.js
+++ b/src/components/EditableList/withEditHandlers.js
@@ -2,6 +2,7 @@ import { connect } from 'react-redux';
 import uuid from 'uuid';
 import {
   compose,
+  defaultProps,
   withHandlers,
 } from 'recompose';
 import { get } from 'lodash';
@@ -49,9 +50,10 @@ const handlers = withHandlers({
     upsert,
     setEditField,
     editField,
+    normalize,
   }) =>
     (value) => {
-      upsert(editField, value);
+      upsert(editField, normalize(value));
       setImmediate(() => {
         setEditField();
       });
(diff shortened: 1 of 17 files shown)


rebeccamadsen:
I think from an NC standpoint `layoutVariable` and `groupVariable` can be null or "" and be okay. I think `layoutVariable from a UX standpoint should be required, though, because narrative is meaningless without a layout.

Behaviours looks like it is doubly nested. Should be:

```
"behaviours" : {
   "freeDraw": true
}
```

I'm seeing this instead:
```
"behaviours" : {
   "behaviours": {
      "freeDraw": true
   }
}
```

Otherwise I think it looks good!

wwqrd:
> I think from an NC standpoint layoutVariable and groupVariable can be null or "" and be okay. I think `layoutVariable from a UX standpoint should be required, though, because narrative is meaningless without a layout.

Would it be better to omit it if left blank?

rebeccamadsen:
> Would it be better to omit it if left blank?

Omitting for `groupVariable` is better than blank; technically the interface in NC is okay with either, but the schema is looking for a string (not `null`) if `groupVariable` is present, which makes sense to me. It looks like `layoutVariable` is required already, which is correct. Otherwise looks great.

PR #410, opened 2019-03-11
Support for editing Per Alter Interface

### LINKED ISSUE #288: Per alter form interface configuration

wwqrd:
## Known issues
- [x] Needs preview image for insert stage screen
- [ ] Description text for insert stage screen might need refining.

Resolves #288


src/components/StageEditor/Interfaces/PerAlterForm.js
```diff
@@ -0,0 +1,15 @@
+import {
+  Name,
+  NodeType,
+  Form,
+} from '../sections';
+
+const NameGenerator = {
```
rebeccamadsen: Should this be PerAlterForm instead of NameGenerator?

PR #412, opened 2019-03-11
Clears the temp directory on app startup

wwqrd:
Clear the temp directory on app start up.

Also took this opportunity to tidy up the filesystem methods.

We should update the redux modules to match (import/export  are now unbundle/bundle).


rebeccamadsen:
Looks okay to me.

PR #431, opened 2019-03-25
Edit forms inline

### LINKED ISSUE #390: Refactor forms
As part of a significant redesign of the Architect workflow, we want to refactor how forms work.

Currently, forms are a top level protocol entity. In the new system, form definitions will be moved inside the stages that use them.

Form definition will use the stage `subject` to lookup variable references.

Proposed format for within stage form definition: 

```
{
  // rest of stage definition...
  "form": {
      "title": "My name generator form",
      "fields": [
      {
          "variable": "2377af3f-3c79-41da-9b0b-6570fb519b93",
          "component": "Text",
          "prompt": "What is this person's name?",
          "validation": {
            "required": true,
            "minLength": 30,
          },
      }
  ]
}
```

Of note:
- The `prompt` property, which mirrors the language used when creating prompts that assign variable values. This is a required property.
- Validation moves to the form field definition.
- **Variables still need to be defined in variableRegistry/codebook, but automatically**

# UI

The component to be changed is `src/components/sections/Form`:
Example interface

The new component will define variables/fields from the interface edit screen (using EditableList):
Form section

Clicking on or creating a new field will show the following edit fields (using EditableList, Editor prop):
Field input

wwqrd:
- Form configuration is now part of particular interface where it is used
- Form fields are added inline, which automatically creates corresponding
  variable entries

Resolves #390


wwqrd:
# Demo/review feedback
- [ ] From the node type editor don't require minimum number of variables.
- [ ] Add /check guidance / descriptive text for new fields
- [ ] For form section, clearly link the title and fields: mini divider? box outline?
- [ ] scrolling with select react component not following, create new issue?
- [ ] TextArea max width (see field prompt), min-height, create new issue?
- [ ] "Prompt" should be a Text input
- [ ] Radio groups etc, should be able to add options
- [ ] Variable name isn't showing up in field editor for previously saved field.
- [ ] Field preview should be improved: variable name (e.g. "Name"), variable type (as color). Perhaps as a "tag" style.

wwqrd:
Text area issue: https://github.com/codaco/Network-Canvas-UI/issues/101

wwqrd:
React select issue: https://github.com/codaco/Architect/issues/433

wwqrd:
This looks done to me? @rebeccamadsen @jthrilly

I think the Options input (for radio group etc) could do with validations, but since that's a pre-existing component, I suggest we make that a separate PR.

PR #441, opened 2019-04-08
Feature/input icons

jthrilly:
This PR updates the icons for the input component chooser.


wwqrd:
Looks good, seems like we have a bit of duplication going on in the input definitions (existed prior to this change)? I'd like to see if I can remove that before we merge this.

PR #442, opened 2019-04-08
Inline node type editing

### LINKED ISSUE #418: Inline node type editing
Update NodeType component and the TypeEditor to create node types inline.

### Node type selector
Nodes can be selected from existing, or created by clicking the new button. A display variable can be set here, and if it is set, it changes how this node is displayed globally (in the registry).
NodeType component

### New node screen/modal
Adding a new node has reduced options, as variables are now defined as they are used.
New node type

wwqrd:
Adds the ablity to create a node type inline, and also select the display variable (though that functionality is only useful on an existing node with existing variables).

Part of: https://github.com/codaco/Architect/issues/417
Resolves #418


wwqrd:
> - the plus icon is off centre. I think it would be better to just continue the pattern from elsewhere of having a "create new +" button in green
> - the label doesn't show up. Did we remove label and just have "name"?

PR #443, opened 2019-04-09
Persist guidance open/close state between screens

### LINKED ISSUE #159: Guidance open/close state should persist between screens
This should be a global property in the redux store.

wwqrd:
Resolves #159


jthrilly:
Can this be added to the rehydrate whitelist so it also persists between app starts? Otherwise looks good!

PR #445, opened 2019-04-10
add CSV file type to network import

jthrilly:
Adds the ability to import CSV files as networks. Translation for CSV data into network JSON is handled within network canvas.


wwqrd:
This looks good to me.

Tangentally: I've noticed a general issue RE display labels for this imported data - it should be driven by the types defined in the codebook (so isn't dependent on the external data), but doesn't seem to show any options. Since the field is required, it can't be saved either:

Screenshot 2019-04-11 at 17 21 44

PR #446, opened 2019-04-11
Feature/updated attributes table ui

### LINKED ISSUE #419: Update attributes table UI
This needs to allow you to create variables inline:

Attributes table

wwqrd:
Replace the AttributesTable with improved AssignAttributes component.

- Allows new variable to be created from select variable dropdown
- Shows correct input types for different variables

Resolves #419


rebeccamadsen:
The attributes work as I'd expect! I'm still not getting protocols from architect to NC, but I think they are unrelated issues. (e.g. `Protocol should NOT have additional properties 'forms'` and `Protocol.stages[1].prompts[0].additionalAttributes should be object` and `protocol.stages[1].prompts[0].additionalAttributes: One or more sortable properties not defined in codebook: 0` and `Protocol.codebook.node['cf0815f4-522d-4b96-8ff6-00ebe431b0d6'] should have required property 'label'`).

Will "quick add" and "display variable" eventually allow variable creation as well?

wwqrd:
@rebeccamadsen actually those features were already added in https://github.com/codaco/Architect/pull/442, but seem to have gone awol, I'll have a look into it.

wwqrd:
Okay, so it's part of quick add (just needed a rebase), but not part of display variable - I'm in two minds about whether it should be since it would allow the creation of 'unused' variables? It's a global setting, so setting it anywhere will update it throughout the app - but that's not really clear either?

PR #449, opened 2019-04-15
Feature/edit ego stage

wwqrd:
Adds support for editing ego interfaces.

Changes a lot of backend things to allow for subject format: `{ entity: 'ego' }` vs `{ entity: 'node', type: '1234-1234' }`

Resolves https://github.com/codaco/Architect/issues/289


jthrilly:
There may be other issues, but I can't test further until these issues are resolved.

jthrilly:
One other small thing to change: the text box for the intro panel text - it has no background so just looks like an offset and incorrectly rendered text box.

wwqrd:
> One other small thing to change: the text box for the intro panel text - it has no background so just looks like an offset and incorrectly rendered text box.

Do you have any direction on how this should look? It's a Markdown input for context

wwqrd:
Ego interfaces should now validate, and you should be able to add variables without any issues.

jthrilly:
In terms of the input, I think just a text area for now. Don't delete the markdown preview component, because I do want to get back to it - it just needs reworking.

For reference, the reworking would probably be something like: a text area with a small "preview" toggle, that replaces the text area with the rendered markdown.

Issue #447, opened 2019-04-12
Some interfaces can't set variables if none are defined


wwqrd:
Some interfaces have required fields which can't be completed unless variables already exist. Currently there is no way to create those variables inline.

Need to add the ability to define these in a modal.

Problem interfaces:
- [x] Categorical bin -> prompt
- [ ] Name Generator List -> prompt: display variable
- [ ] Name Generator Search -> prompt: search properties, display variable
- [x] Narrative -> preset: layout variable, group variable
- [x] Ordinal bin -> prompt

wwqrd:
We're going to read external data for the rest of these properties

wwqrd:
Issue for the external data properties: https://github.com/codaco/Architect/issues/453

PR #451, opened 2019-04-15
Fix/variables must always be "creatable"

### LINKED ISSUE #450: Inline deletion of variables
We need a way to edit variables.

Deleting variables inline is already a wanted feature, and will allow a user to delete and then recreate variables as a stop-gap to editing.

wwqrd:
Adds inline variable creation for the following:
- "Categorical bin -> prompt"
- Narrative -> preset: layout variable, group variable
- Ordinal bin -> prompt
- Name Generator Search -> prompt: search properties

Partially completes https://github.com/codaco/Architect/issues/447

Resolves #450 (inline variable deletion)


wwqrd:
"Name Generator Search -> prompt: search properties" and the display labels from  #447 are questionable features, so I suggest we leave this branch as is for now (and merge it).

PR #455, opened 2019-04-20
Alter edge form interface editing

### LINKED ISSUE #439: Alter edge form interface editing
Ability to edit the alter edge form interface.

wwqrd:
Adds capability to edit Alter Edge form interfaces.

Also refactors/fixes "subject is required" sections.

Resolves #439

### DIFF
--- a/src/components/StageEditor/Interfaces/AlterEdgeForm.js
+++ b/src/components/StageEditor/Interfaces/AlterEdgeForm.js
@@ -1,0 +1,21 @@
+import {
+  Name,
+  EdgeType,
+  Form,
+  IntroductionPanel,
+} from '../../sections';
+
+const template = {
+};
+
+const AlterEdgeForm = {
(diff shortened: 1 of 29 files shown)


rebeccamadsen:
These could be related, but this is what I noticed:

When adding a field to a form, "Save and continue" button does not close the dialog.

It looks like the variables I add in a field are not saved to the codebook -- both in name generator and in alter edge interfaces. I can't use one of the new variables as a display variable, or in a filter rule on a panel. If I click back on a variable to edit it, the variable name is missing.

The subject on the interfaces seem to be missing the "entity".

The structure looks good otherwise.

wwqrd:
I can't replicate this?

- Are you running in electron or browser?
- Which protocol are you editing?
- What is the commit sha?

wwqrd:
Have replicate with a new protocol in electron

PR #462, opened 2019-04-24
Fix/small style changes

jthrilly:
A few small style changes:

- Made the start screen more responsive and added file path (at the expense of the flip animation)
- Harmonized the stage-editor-section styling to use the styles from the form section, because I think they look great.
- Updated the UI ref, which fixed an issue with the Toggle component label.
- Changed the order of the sections in various screens
- Provided updated guidance copy in a few places
- Updated the "view code" button to be an icon, and made it inline with the section heading.


jthrilly:
To do: clone new stage button so that it can be large on timeline only.

jthrilly:
I decided to replace the `<NewButton>` with `<Button>` since button also allows text to indicate the action, and is used in the node type selector (so this is more consistent).

PR #466, opened 2019-04-26
Feature/use shared development protocol

### LINKED ISSUE #377: Mock data in development mode should be driven by shared development protocol submodule
~File actions are mocked when developing architect on the **web** development mode. At the moment this is a custom file `mockProtocol.json`, it should be updated to read from a submodule which is shared between Network Canvas and Architect.~

Retire 'web' support in Architect (`npm run start`). We should still support: (`npm run start:electron` and `npm run electron:dev`)

wwqrd:
This is an attempt to resolve #377, by including the development protocol as a submodule.

It follows the existing pattern of importing the protocol file but has the following drawbacks:
- assets are still not usable
- external data (csv) is still loaded from text inline
- external data (json) is loaded from a hard-coded reference, so always references the same file.

#### Include development protocol in public, and write XMLHttpRequest shims to fetch data

The drawbacks with this are:
- will need to write code to emulate file loading over http
- will need to ensure this code is not included in production build
    
#### Retire web mode and rely on desktop app (which can load the development protocol from a file)

The drawbacks with this are:
- lose the benefits of fast iteration, when the protocol requires changes.

Resolves #377

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,2 +1,3 @@
+src/development-protocol
 src/ui
 src/network-canvas
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "src/network-canvas"]
 	path = src/network-canvas
 	url = https://github.com/codaco/Network-Canvas
+[submodule "src/development-protocol"]
+	path = src/development-protocol
+	url = git@github.com:codaco/development-protocol.git
--- a/package.json
+++ b/package.json
@@ -205,6 +205,7 @@
     "jest": {
         "collectCoverageFrom": [
             "src/**/*.{js,jsx}",
+            "!src/development-protocol/**",
             "!src/ui/**",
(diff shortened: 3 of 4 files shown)


wwqrd:
@jthrilly and I think retiring this feature makes the most sense.

jthrilly:
So, perhaps I'm missing something, but is there some reason not to just manually open the development protocol? Why do we need a workflow for it?

wwqrd:
> So, perhaps I'm missing something, but is there some reason not to just manually open the development protocol? Why do we need a workflow for it?

Yep, that's the second option. Discontinue support for web, and just do development in electron (which will allow manually opening the development protocol)

jthrilly:
Ah, great. Then that's what I think it's best. In terms of things to change, I think validation would be the thing to include as a submodule, and then implementing that on file open and file save.

PR #468, opened 2019-04-30
Improved content items

### LINKED ISSUE #358: Improve information interface editor
There are some usability issues with the content boxes on the information interface.

User feedback:

> I pressed the add content button, but then changed my mind because I didn’t have a picture readily available. There is no apparent way to cancel out of that action without canceling everything else on that page (name, page heading, etc.) and starting over. 

- [ ] Add a cancel button next to the done button

> There is also no way to switch between different content types (i.e. if I wanted to initially add an image, but changed my mind and decided I wanted to add text instead) without adding a new box. 

- [ ] Either add the ability to switch types (make the type chooser a radio group?) or this can be solved by cancelling and adding another box.

> Content is required for each box I initiate if I want to continue, even if I’ve changed my mind. Is there a way we could make this more flexible?

This will be resolved by a combination of the above.

wwqrd:
Slightly indirect fix for #358 

Replaces ContentGrid implementation with one based on `react-grid-layout`.

- Item editing happens in a modal
- Items are now dragged to size, rather than using a mode selector

Resolves #358


wwqrd:
I think this is good to go now

PR #480, opened 2019-05-13
Preview mode 2.0

### LINKED ISSUE #373: Preview for ordinal bins doesn't match NC
Just noting this, should be resolved by: https://github.com/codaco/Architect/issues/361

screenshot 2019-01-21 at 10 22 29
screenshot 2019-01-21 at 10 22 36

wwqrd:
New implementation of preview which defers Network Canvas build to itself and which is then packaged up with Architect build in `/app/`.

Controls Network Canvas through electron IPC.

- Move build into `app/` rather than `www`
- Remove embedded preview components
- Move network-canvas submodule to top-level of app
- Add npm tasks that build network canvas
- New preview manager that defers to an independent build of network canvas packaged with the app.

Requires: https://github.com/codaco/Network-Canvas/pull/903

Resolves: https://github.com/codaco/Architect/issues/361, Resolves #373

**Note: this is a duplicate of https://github.com/codaco/Architect/pull/478, which was reverted after a erroneous merge**


rebeccamadsen:
Preview is looking better! There was a weird moment just now that scrolling wasn't working in architect, but it did work before. I'll try it again tomorrow and see if I can reproduce. I could still use my touchscreen to scroll though?

Assets don't work for preview (like background image on sociogram, csv files or previousInterview nodes, audio, video). Maybe that's by design though. The interface shows, just not the asset. I could add nodes though! And see them on another interface! Menu doesn't show when clicked (again, by design, I'm thinking), but I could navigate stages, which was fancier than I expected.

rebeccamadsen:
I can't duplicate the scrolling issue I saw last night, but I also can't see previews today. Not sure what I'm doing differently. I tried to create a build, thinking that might be better and see this error:
```
npm ERR! path C:\Users\Big Baloo\Documents\GitHub\Architect\network-canvas\build
npm ERR! code EISDIR
npm ERR! errno -4068
npm ERR! syscall open
npm ERR! EISDIR: illegal operation on a directory, open 'C:\Users\Big Baloo\Documents\GitHub\Architect\network-canvas\build'
```

jthrilly:
Only issue for me is that the finish screen should not show - everything else works as expected.

wwqrd:
> Assets don't work for preview (like background image on sociogram, csv files or previousInterview nodes, audio, video)

I'm thinking that this may be a platform thing? Would you be able to investigate and fix? It's probably an issue with: https://github.com/codaco/Architect/blob/4a93ce26b6df3aca328f7744c37df21e7cbfb6da/public/components/assetProtocol.js#L16

wwqrd:
@rebeccamadsen, hopefully fixed this for windows, will need to update submodules to check:

In Architect:
- Build urls with `asset://${systemPath}`, in unix this means we get paths like `asset:///foo`, but also:
- In asset protocol remove full `asset://`, which brings us back to the full path. This (in theory) will make it work correctly in windows.

In Network Canvas:
- use `encodeURIComponent()` and `path.join` in electron, this should ensure we get the correct slashes for the system, and anything else we need is encoded properly.

rebeccamadsen:
That fixes for windows! I can see assets in both the preview and in architect! This is exciting.

There is one (small) weirdness with the skip logic interface for the development protocol, now that we are hiding the finish screen for preview mode. You see the skip logic interface as normal when you preview it directly, but if you click the next button inside preview mode until you get to the interface, the skip logic takes effect. Then it tries to "skip" to stage 22, but that stage doesn't exist (this is the finish screen in NC) so we see "No "" interface found." I don't think this is a big deal, though, and could be addressed in a future issue, since the interface shows as you would expect if you are previewing each interface individually. It also is only an issue here because the skip logic interface in development protocol is at the end; a skipped stage anywhere else in the flow wouldn't have this issue.

PR #481, opened 2019-05-14
Unique names for variables

### LINKED ISSUE #471: Variable names must be unique
Variable names must be unique. When creating a new variable (through the inline select, or through the modal) this must be enforced.

wwqrd:
Adds field validation for variable creation that checks against existing variable names (case insensitive).

RE enforcing this for the creatable option type: in my testing this already wasn't possible  as the existing variable will be selected. We can use the `isValidNewOption` prop if that turns out to be incorrect (https://react-select.com/props#creatable-props).

Resolves #471

### DIFF
--- a/src/components/EditableList/EditableList.js
+++ b/src/components/EditableList/EditableList.js
@@ -52,6 +52,7 @@ class EditableList extends PureComponent {
       initialValues,
       editComponent: EditComponent,
       previewComponent: PreviewComponent,
+      editProps,
       ...rest
     } = this.props;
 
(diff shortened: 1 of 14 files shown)


rebeccamadsen:
I see the error when I try to add a variable name that was already used, which is great. However, when I try to open a form field later to edit it, the variable name (and options, if I had them) are missing. In NC I get `Error: Protocol.stages[1].form.fields[0] should have required property 'variable'` for the protocol.

jthrilly:
I confirmed @rebeccamadsen's bug, and am fixing it now. Will merge this once I do

wwqrd:
Might also be worth resolving here: quickAdd is not removed when switching between the two types of prompt - although this issue will be resolved by the proposed pseudo interface change: present the two quick add/fields configurations as a non-interchangable preselected option.

Issue #475, opened 2019-05-07
Removing an asset does not remove references to it


jthrilly:
1 - Add an image asset
2 - Add as background image to one or more sociogram screens
3 - Remove image asset from asset manager
4 - Asset is still referenced by UID in sociogram configuration, but the editor shows concentric circles selected.

Solution: do not allow removing assets that are in use. Inform user where asset is in use when they attempt to delete.

wwqrd:
Adding an asset through the asset manager is not a savable change either, suggest this is fixed here.

wwqrd:
Also might be worth reviewing how unused assets are deleted from disk, to use the same code paths as the usage checker here.

PR #482, opened 2019-05-15
Prevent deleting used assets

wwqrd:
This updates asset browser UI as follows:

  - Show not-allowed cursor, and tooltip text for in use assets on delete button
  - If delete button is clicked, show further explanation in an info dialog
  - Show an "unused" tag on unused assets.

### Known issues
Messages are generic, i.e. do not explain where usage is occuring. That information is available in the backend (in the form of a path e.g. `stages[1].prompts[2].dataSource`), considering that this info is quite awkward to explain perhaps we could give a top-level notice about which screens it is used on.

Resolves: #475

### DIFF
--- a/src/__tests__/testState.json
+++ b/src/__tests__/testState.json
@@ -13,6 +13,18 @@
       "assetManifest": {
         "1234-asset-1": {
           "source": "1234-4567-asset-1"
+        },
+        "1234-asset-2": {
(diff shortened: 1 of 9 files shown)


rebeccamadsen:
Works well for assets used in information and sociogram and rosters. The network data file I added to a side panel in name generator, though, I could still delete, which errors out in NC because the asset is missing.

PR #484, opened 2019-05-16
Release/interim build 2

### LINKED ISSUE #387: Rename "additional attributes" to "Assign attributes"
The additional attributes section of the prompt on the name generator should be renamed to "assign attributes", and include some short text on its purpose.

This is a UI change, not a protocol change (so `additionalAttributes`) can be retained in the protocol file. 

This will resolve #359

jthrilly:
Another round of feedback:

- Added text for additional attributes. Closes #387.
- Updated the sociogram prompt text
- Moved sort options, card options, and search options (`NameGenList` and `NameGenAutoComplete`) to the stage level
- Removed showExistingNodes
- Added `NameGeneratorQuickAdd` interface type.
- Updated filtering to use `attribute` rather than `variable` (consistent with networkQuery in use in NC and Server)


jthrilly:
One further task before this is ready:

```
bg2.netcanvas has schema errors:
- Protocol.stages[0].skipLogic.action should be string 


- Protocol.stages[0].skipLogic.action should be equal to one of the allowed values 'SHOW,SKIP' 
```

The skip logic function is producing the wrong format.

PR #487, opened 2019-05-21
Fix/sort options

### LINKED ISSUE #474: No sort properties available in sociogram prompt (elsewhere?)
The sort properties for the node bin in the sociogram prompt configuration is not showing any node properties.

wwqrd:
Fixes sort options for the sociogram layout

Checked for issues elsewhere and followed a red herring re external data: due to improperly formatted json the external properties weren't being displayed. On the plus side, I renormalized the enhancers during my attempt to 'fix' the issues (took them out of Prompts where they are no longer used).

Resolves #474


wwqrd:
Also raises the question of checking json/csv validity import:
- json that fails to parse will be silently imported, only to mysteriously not show expected properties in some of the interface editors.
- json is assumed to always be in a network format.
- csvs are also imported without checking content

PR #496, opened 2019-05-31
Add save copy feature

wwqrd:
- Also revises electron app structure
- Adds 'save' button in menu
- Updates confirm close to include draft stage edits


rebeccamadsen:
The saving and copying looks great! I can save files, and save copies, and reopen them how I would expect. 

With respect to file associations, double clicking on a .netcanvas file opens Architect; I think it opens the Architect version in `C:\Users\Big Baloo\AppData\Local\Programs\network-canvas-architect`, as near as I can tell -- my last installed version. With an release install created from this branch, however, it seems to only open Architect and doesn't open the protocol file I clicked on. Double-clicking a .netcanvas file while Architect is already open gives me a couple of errors and also does not open the protocol for me:

arch

arch2

wwqrd:
@rebeccamadsen I think I've fixed this for Windows, would you please confirm?

I think windows file association is triggered by the app installer, so that's why the old version was loading initially. Rather than a full instally, another way you can test it is to set the association up manually in file explorer directly with the exe.

PR #497, opened 2019-05-31
fix namegen form field preview

jthrilly:
Fixes name generator form field previews, which had stopped displaying variable type and component


rebeccamadsen:
Looks good to me.

PR #498, opened 2019-05-31
When edges are created, make sure they are visible

wwqrd:
When a prompt is configured to create an edge type, automatically select that edge type for display (and prevent untoggling).

Resolves: https://github.com/codaco/Network-Canvas/issues/760


rebeccamadsen:
Looks good!

PR #503, opened 2019-06-04
make component a required variable property for select option on name…

### LINKED ISSUE #501: reusing ordinal/categorical variable should show options
Reusing an ordinal or categorical variable is possible, but we should also show the existing options for the variable. Right now, if a user wants to reuse an ordinal or categorical variable, they will need to create new options to save the field, which will overwrite any previous options. We should show them what they've already created.

Changing to a different type is possible, but probably should reset options.

jthrilly:
Fixes #501 

From Slack:

>Having thought about this a bit more, I think what we actually need to do is to exclude variables that do not have an input type (and were therefore not defined on a node form) from the select box for node form variables.
>
>Reason is simple: if I create an ordinal variable on an ordinal bin (without an input type), and then add it to a name generator form, and then _change_ the type, it will invalidate the ordinal bin interface entirely.
>
>For the same reason, I think the ordinal bin interface should exclude ordinal variables that _have an input type_ from its select box.

Issues:

- causes a crash when the participant types an existing variable name (that is filtered from the list because it doesn't have a component) and presses enter.
- doesn't work for creating a new variable (which of course has no input component set yet!)

### DIFF
--- a/src/components/sections/CategoricalBinPrompts/withPromptProps.js
+++ b/src/components/sections/CategoricalBinPrompts/withPromptProps.js
@@ -1,11 +1,23 @@
 import { connect } from 'react-redux';
 import { change } from 'redux-form';
 import { compose, withHandlers } from 'recompose';
-import { getVariableOptionsForSubject } from '../../../selectors/codebook';
+import { pickBy, reduce } from 'lodash';
+import { getVariablesForSubject } from '../../../selectors/codebook';
 import { actionCreators as codebookActions } from '../../../ducks/modules/protocol/codebook';
 
 const mapStateToProps = (state, { type, entity }) => {
-  const variableOptions = getVariableOptionsForSubject(state, { type, entity });
+  const existingVariables = getVariablesForSubject(state, { type, entity });
+
+  const variablesWithoutComponents = pickBy(existingVariables, value => !value.component);
+
+  const variableOptions = reduce(
+    variablesWithoutComponents,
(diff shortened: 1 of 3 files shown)


jthrilly:
I don't like this approach: 

I think Steve was right earlier when he said we should look at making type static once set. I imagine it would go like this:

- When creating a *new variable*, there should be an additional input for choosing the variable type, directly following the name.
- Type should be used to filter the available options for input component, which would then show/hide the options section as normal.
- Once saved, type would be permanently set, so selecting an existing variable would result in this field being in the disabled state.

We need another feature to go along with this, that is also currently missing.

- On the ordinal and categorical bin interfaces, when you create a variable and associated options, there is no way to edit (or even view) them.
- We can solve this by reusing a version of the modal for the name generator form fields, except removing the input control, validation, and the question prompt fields. This could be triggered by a simple button under the variable selector that says "edit variable", OR when creating a variable from that select.

jthrilly:
Third attempt at solving this (agreed on the call just now:

- Use `type` instead of `component` to determine if options are shown when a variable is selected. Type is available on existing variables that were created elsewhere.
- New variables will have type inferred by component as it currently works. 
- If type is present, the component options will be filtered by that type. This will make type permanent, so the user cannot accidentally invalidate other interfaces.

wwqrd:
Additonally:
- `options` and `validations` should only reset when type changes, not when input component changes.

PR #505, opened 2019-06-05
Reinstate codebook view for deletion of unused types/edges

wwqrd:
Screenshot 2019-06-05 at 15 29 16


rebeccamadsen:
Works great for removing unused Node Types!

Two things for edges:
I can delete an edge that is listed in the Narrative interface, as long as it isn't used in the sociogram I created it for. It is still in the protocol file for that Narrative stage, though this doesn't error in NC, and it doesn't show up in the ui in Architect for that stage.

I can create an edge type in Per Alter Edge Form, and it shows as unused in the "manage codebook". Although, attempting to save a protocol with an edge that was only used in Per Alter Edge Form doesn't make sense (there will be no edges in any interview because this stage doesn't create edges), and it also has an error about "iconVariants", so I can't save it anyway. But it isn't an unused edge, technically?

jthrilly:
Great catches Rebecca. We are going to update this function to address these non-typical locations for edge references: https://github.com/codaco/Architect/blob/3b26f07d6e2a0fb8ebeb69a8f558aeba6f549980/src/selectors/indexes.js#L13

rebeccamadsen:
Not sure if this is done yet or not, but what I've tested now seems to work well!

Issue #506, opened 2019-06-05
error using "0" for values


rebeccamadsen:
Here are some examples of where "0" is not an accepted value in the ui, though it should be allowed.

1. Skip logic using a number variable:
![skip logic 3](https://user-images.githubusercontent.com/16093053/58972853-84852400-878c-11e9-88d5-c3f745cb70af.png)

2. Ordinal variable creation on ordinal bin (categorical bin is similar):
![ord](https://user-images.githubusercontent.com/16093053/58972911-a67ea680-878c-11e9-95f6-248524a593b7.png)

wwqrd:
Resolved by https://github.com/codaco/Architect/pull/511

PR #511, opened 2019-06-05
Fix skip logic preview and falsy value

wwqrd:
- Use toggle for boolean value, which auto sets to false on mount.
- Update field validation to accept boolean values, or 0

Resolves https://github.com/codaco/Architect/issues/508


rebeccamadsen:
This fixes the ui for me, and the false value doesn't have to toggle to be false. However, it shows "required" when the value is false in the form:

false

I can still Save though, so I'm not sure what the complaint is there.

wwqrd:
I updated the ‘required’ validation to resolve but removed when I saw some overlap with another validator. Will check back on this when I know

wwqrd:
Okay, I've update the validation rule. Along with another bug I found when using the number input.

Should also resolve #506

src/utils/validations.js
```diff
@@ -13,9 +13,17 @@ const coerceArray = (value) => {
...
+
+  return !isNil(value);
+};
+
 export const required = () =>
   value =>
-    (value ? undefined : 'Required');
+    (hasValue(value) ? undefined : 'Required');
 
 export const requiredAcceptsZero = () =>
```
jthrilly: Should this be removed, now?

PR #516, opened 2019-06-06
Fix/attributes table value setting

### LINKED ISSUE #513: require values for additionalAttributes
Related to #264. We made additional Attributes default boolean values to true before, and it should do so again. Otherwise, no value is recorded.

Also, it is possible to create variables of other types (text, ordinal, etc) and save the prompt without assigning a value to them. Those will generate protocol validation errors when trying to save the protocol later. Can we require a value if a variable is selected on additional attributes? A variable is optional, but once a variable is there, the value should be required.

wwqrd:
- Prevent variable from being changed once set (user can instead delete it an add another)
- Require value property

Resolves #513

### DIFF
--- a/src/components/AssignAttributes/Attribute.js
+++ b/src/components/AssignAttributes/Attribute.js
@@ -33,6 +33,7 @@ const Attribute = ({
           component={Select}
           options={variableOptions}
           onCreateNew={() => { onCreateNew(index); }}
+          isDisabled={!!variableType}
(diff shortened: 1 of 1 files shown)


rebeccamadsen:
Works for me, although "0" for number or ordinal variables triggers the "Required" warning, as does a false value for boolean variables. Those are fixed in master, though, right?

PR #523, opened 2019-06-10
Ensure variable names are compatible with server/nedb

wwqrd:
Updates variable creators to disallow the following characters:
```
. $ [ ] { }
```

All other characters are allowed (inc. space).


wwqrd:
TODO: Let's re-run this check in the action dispatcher, to make 100% sure it is enforced.

PR #524, opened 2019-06-10
Fix/narrative categorical variable

rebeccamadsen:
This fixes a typo, and makes sure the categorical variable selection for Narrative interface gets added to the stage correctly. 

It also makes the group variable on Narrative a regular `Select` instead of a `CreatableSelect`.


wwqrd:
Looks great to me

PR #546, opened 2019-06-17
Update silently

wwqrd:
Partially resolves https://github.com/codaco/Network-Canvas/issues/929


wwqrd:
Tested locally but not verified as working.

PR #567, opened 2019-08-07
Allow ordinal and categorical vars to be used in skip logic

### LINKED ISSUE #483: Ordinal and categorical variables cannot be used in filters/skiplogic
As in title: these variable types cannot be used in filers.

Implementation of this would require:
- [x] Allowing an ego/alter/edge variable rule to choose variables of this type
- [x] Allowing the operator to be either EXACTLY or NOT
- [ ] Rendering a checkboxgroup of possible values, but **only allowing one item to be selected** (multiple rules should be chained together to include multiple values)
- [ ] The network query lib should be updated to implement checking within the array values that these variable types are stored as

wwqrd:
> Rendering a checkboxgroup of possible values, but only allowing one item to be selected (multiple rules should be chained together to include multiple values)

Have updated rule builder to:
For Categorical variables:
- Show a checkbox group which only allows a single value (still stored as array value)
- New predicate types: includes and excludes
  - technically this work on an any/none basis, but because we can only select one value this functionality is disabled in use.
- Ordinal shows a radio group and allows 'exactly' type matches

Partially resolves #483


wwqrd:
Should we reset recentProtocols, when app schema version changes?

jthrilly:
> Should we reset recentProtocols, when app schema version changes?

Not a bad idea!

jthrilly:
I opened an issue to cover one last potential issue with stage level node filtering: https://github.com/codaco/Architect/issues/583

Could also be resolved here if we have the desire. It hinges on if the current behaviour (allowing filter objects without a `join`) breaks NC. This needs testing as part of this PR

PR #573, opened 2019-08-23
update electron

rebeccamadsen:
Updates electron to 5.0.9, node to 12.8.0, and npm to 6.10.2. Will need to run `npm install`. Note: this does not yet update react, cordova, or the various other dependencies, so we can separate concerns in the updating process.

Will need an updated reference to NC when that is merged in.


package.json
```diff
@@ -106,11 +105,11 @@
         "jsdoc": "^3.4.3",
         "jsdoc-babel": "^0.3.0",
         "json-loader": "^0.5.7",
-        "jszip": "^3.1.5",
+        "jszip": "^3.2.2",
         "lodash": "^4.17.11",
-        "mdns": "^2.4.0",
-        "mini-css-extract-plugin": "^0.4.3",
-        "node-sass": "^4.9.3",
+        "mdns": "^2.5.1",
```
jthrilly: Was interested to see this - where does Architect get its need for this from?

public/components/createPreviewWindow.js
```diff
@@ -9,6 +9,9 @@ const windowParameters = {
   show: false,
   width: 1024,
   height: 768,
+  webPreferences: {
+    nodeIntegration: true,
```
jthrilly: As per my comment in the Network Canvas PR, it would be nice to be able to lock this down.

src/other/protocols/importAsset.js
```diff
@@ -28,8 +28,7 @@ const getTypeFromMime = (mime, mimeTypes = MIME_TYPES) => {
 };
 
 const getTypeFromExtension = (name, extensionTypes = EXTENSION_TYPES) => {
-  const extension = path.extname(name);
-
+  const extension = toLower(path.extname(name).substring(1));
```
jthrilly: Did this throw up a compile/linting issue, or was it just a bug you found and also fixed?

src/other/protocols/importAsset.js
```diff
@@ -103,7 +102,7 @@ const validateCsv = data =>
 export const validateAsset = (file) => {
   const assetType = getTypeFromMime(file.type) || getTypeFromExtension(file.name);
 
-  if (assetType !== 'network') { Promise.resolve(true); }
+  if (assetType !== 'network') { return Promise.resolve(true); }
```
jthrilly: Same question with this...was it a bug?!

jthrilly:
As with the other PR, I can't get this one to launch. The window opens and immediately closes, with the last log item being `[updated]`. Looking into it a little more now.

jthrilly:
> As with the other PR, I can't get this one to launch. The window opens and immediately closes, with the last log item being `[updated]`. Looking into it a little more now.

Looks like the check for the single instance lock is returning false every time on my system. Looking into it now.

jthrilly:
Looks like my app wasn't getting the single instance lock. I checked and there were no other processes running with that name, and a reboot fixed it (...!). I think we should test that this doesn't cause an issue with updating.

jthrilly:
This also seems to suffer from the menuItem issue that is impacting the NC version. Whatever the fix is there should also work here since the menu construction is the same.

jthrilly:
Preview mode seems to not be operational at all (events fire but the preview window doesn't open).

src/other/protocols/importAsset.js
rebeccamadsen: Possibly a bug; I haven't checked master (yet). I couldn't add image assets before I added this.

src/other/protocols/importAsset.js
rebeccamadsen: Possibly a bug. I had a file with a .JPG extension that wasn't being recognized as an image asset; it was being handled as a network file.

src/other/protocols/importAsset.js
wwqrd: Nice spot! If not a bug now, one waiting to happen! Can we add a test?

PR #574, opened 2019-09-10
fix accept list for dragging on windows

### LINKED ISSUE #476: CSV file handling on windows is broken
On Windows there are issues identifying CSV files with our drag and drop lib: https://github.com/react-dropzone/react-dropzone/issues/276

We were getting an empty `file.type` in my testing. I bodged around this by using the file extension, which works.

The issue that remains is that I couldn't get the `accepts` property to recognize `.csv` as valid, even though the docs say that you should be able to specify a file extension.

Dragging over a CSV file will show it as invalid, although dropping it will have the file work as normal.

rebeccamadsen:
This fixes #476. Per [this](https://github.com/react-dropzone/react-dropzone/tree/master/examples/accept), "if you want to react on different file types during the drag operation, you have to use mime types and not extensions". Csv files in windows use `application/vnd.ms-excel` as the mime type.


wwqrd:
Naive question from non windows user: Is this mime type sent even when you don't have office installed? i.e. should we remove the existing workaround?

jthrilly:
> Naive question from non windows user: Is this mime type sent even when you don't have office installed? i.e. should we remove the existing workaround?

The situation is extremely complicated. Apparently different OS and Office versions change the MIME type for CSV. As far as I know, there's no reliable way of detecting it (I could have done what this PR does at the initial time of noticing it, but I didn't because of exactly this). 

Looking at it with fresh eyes, perhaps it is better to have this work some of the time, for at least some users. I'm inclined to have this in and see how it goes, since it is generally an improvement to the situation.

wwqrd:
I'm definitely for the addition, I'm just wondering if we should also be removing 'text/*', unless it's not actually doing anything?

jthrilly:
> I'm definitely for the addition, I'm just wondering if we should also be removing 'text/*', unless it's not actually doing anything?

Yeah, I don't think we should. Good spot.

I added it because of this SO answer: https://stackoverflow.com/questions/7076042/what-mime-type-should-i-use-for-csv#answer-42140178

Basically, there are a whole host of `text/` mime types that might be sent, and `text/csv` is the standard that _should_ be sent (and is on everything but windows). Chrome is deferring to the Windows registry at the moment, but could one day be updated to support the correct RFC. 

I'll add the text rule back, which certainly won't hurt.

rebeccamadsen:
The text/* rule allowed me to a add .txt file as a network file, even though it was just plain text in the document. Saving the protocol went fine, but "previewing" or taking the protocol to NC means that NG stage errors out. I think I may have another computer with windows and not Office I could test out, but not sure if that's sufficient. Do we prefer false negatives or false positives??

wwqrd:
I would have thought false positives were better? The user is unlikely to intentionally sabotage their own protocol.

However, this also sounds like a bug, why isn't validateCsv flagging this?

rebeccamadsen:
Good question! csv-parse turns out just fine. It's a pretty simple text file -- just one line with some text and spaces and punctuation. I added some commas, but it wasn't until I added a new line that csv-parse was unhappy and rejected my file. Hopefully users are submitting real files and not random text files though.

jthrilly:
I reworked this a bit.

- Removed `accepts` entirely, so now we accept any file, and then validate after that.
- Consolidated the logic for accepting. There's now a single function, `getSupportedAssetTytpe`, which returns either the asset type, or false if the asset is unsupported. 
- This function first uses mime type, including Rebeccas additional one, and then falls back to extension. This is needed because mime type is empty for windows computers that don't have office installed.
- Based on this outcome, the user either sees a dialog that the asset type is not supported, or network validation happens. Validation can fail, causing a different dialog.

src/ducks/modules/protocol/utils/dialogs.js
```diff
@@ -26,7 +25,30 @@ export const invalidAssetErrorDialog = (e, filePath) => {
...
-    error: e,
+    message: friendlyMessage,
+  });
+};
+
+export const unsupportedAssetErrorDialog = () => {
+  const friendlyMessage = (
+    <React.Fragment>
+      <p>
+        The file you attmpted to import is not in a format supported by Network
```
rebeccamadsen: Should be "attempted", not "attmpted".

jthrilly: > This takes away the "type" handling. Adding a "network data file" on a side panel for a "Name generator" interface now allows any valid type of file to be dropped, including images, video, csv, etc.

Good point. I'll think about this a bit more.

jthrilly:
Okay, final set of changes:

- The library now works using file extensions in `accepts` so I was able to reinstate that, which avoids some potential complexity with the name gen panels.
- This meant I could undo the stuff I did yesterday adding in a new dialog for invalid types. I think this is a much better pattern (preventing the user from doing something they can't do, rather than telling them off for doing it).
- This has the added benefit of scoping the native OS "browse" dialog to the types of files that are allowed.

I tested CSV on Windows, both with office installed and without, and it is now working correctly.

wwqrd:
Unrelated to these changes: I'm not sure when the preview for the video/audio assets changed, but it looks odd on the info screen.

Screenshot 2019-09-17 at 11 55 13

PR #575, opened 2019-09-16
Fix for ordinal and categorical prompts not having an ID

jthrilly:
Several users reported protocol breakages when using the last release because prompts created on either the ordinal or categorical bins were not being given an ID. 

I tracked it down to the `itemSelector` helper, which is duplicate code except in the case of creating a new prompt. In the broken code it returned an empty object (truthy), which subsequently meant the default parameters weren't being added (including ID). 

The other version of this code returns null in this case, and this fixes the issue.

Two things I am not sure on:

- Can we not have two implementations of this?
- How could we have caught this automatically through a test? One aspect of it that is interesting is that editing a prompt that already has an ID won't display the broken behavior. This is why we didn't catch it when testing against the development protocol.

Addressing either of these should be done separately, as I need to release this fix today. Discussion on those points welcome, but please just review this PR on the basis of the fix, since it is a priority.


wwqrd:
> - Can we not have two implementations of this?

This method is for injecting custom behaviour into a generic "prompts" component, basically to side-load any other meta related to a prompt. The ordinal and categorical prompts already share this helper function, but other prompts might need to bring in different data.

> - How could we have caught this automatically through a test? One aspect of it that is interesting is that editing a prompt that already has an ID won't display the broken behavior. This is why we didn't catch it when testing against the development protocol.

Since it's just a plain old function we could add a unit test. The ugly part of that would be that we would probably want to stub the `formValueSelector` function. Another way would be to add it as a functional test, once we have something compariable to pending the integrations tests in NC: https://github.com/codaco/Network-Canvas/pull/955

jthrilly:
> Since it's just a plain old function we could add a unit test. The ugly part of that would be that we would probably want to stub the `formValueSelector` function. Another way would be to add it as a functional test, once we have something compariable to pending the integrations tests in NC: [codaco/Network-Canvas#955](https://github.com/codaco/Network-Canvas/pull/955)

Yeah, the obvious one is to make sure that we test creating a prompt from scratch when we compile the list of core tests for that functionality. Good call.

wwqrd:
> - Can we not have two implementations of this?

I've pushed up a change which I think better addresses your point, basically moving the form selector functionality up to the EditableList component, it makes `itemSelector` less flexible, but at the moment it's used the same way everywhere so removing support for this unused "feature" makes the most sense here.

jthrilly:
Looks like a positive change to me. I'll merge and start a release branch as soon as the tests pass.

PR #585, opened 2019-09-25
Fix/variable case sensitivity

### LINKED ISSUE #578: case sensitive variable names
NC treats variable names as case sensitive, but Architect uses case insensitive matching to prevent duplicate names. Case sensitivity handling for variables should match across apps.

For example, a user can't create a variable in Architect named `Name` and then create `name`, because it is considered a duplicate. NC won't recognize `Name` as the preferred label for a node, though, so this means the protocol will have to resort to fallback options for labels.

Consider allowing users to create variables with upper or lower case, and treat them as different variables. For this use case, users could then create a `name` variable.

We should also verify that on export from NC and Server, case sensitivity is also treated the same way.

jthrilly:
This PR implements case sensitivity on variable creation. Resolves #578 

Other changes:

- Updates react to 16.8
- Updates react-select to 3.x
- Changes text accompanying creation to clarify that pressing enter will not always create a variable
- Allows the input to be cleared


src/components/Form/Fields/CreatableSelect.js
```diff
@@ -124,10 +137,22 @@ class Select extends PureComponent {
           // a round about way, and still allow us to use the `touched` property.
           onBlur={this.handleBlur}
           blurInputOnSelect={false}
+          isValidNewOption={(option) => {
+            // Logic for showing the "create" option is simple:
+            // If the input value matches an item in the options collection,
+            // don't show. Otherwise, show.
+            const existingIndex = findIndex(options, o => o.label === option);
+            if (existingIndex > -1) {
```
wwqrd: We could remove the conditional:

```js
(option) => {
  const existingIndex = findIndex(options, o => o.label === option);
  return existingIndex === -1;
}

// or maybe even using some

(option) => {
  const alreadyExists = options.some(({ label }) => label === option);
  const isEmpty = option.length === 0; // We also need to check if the string is empty I think?
  return !isEmpty && !alreadyExists;
}
```

src/components/Form/Fields/CreatableSelect.js
```diff
@@ -124,10 +136,19 @@ class Select extends PureComponent {
           // a round about way, and still allow us to use the `touched` property.
           onBlur={this.handleBlur}
           blurInputOnSelect={false}
+          isValidNewOption={(option) => {
+            // Logic for showing the "create" option is simple:
+            // If the input value matches an item in the options collection,
+            // don't show. Otherwise, show.
+            const alreadyExists = options.some(({ variableLabel }) => variableLabel === option);
```
wwqrd: `variableLabel` should this just be `label`, or at least `{ label: variableLabel }`? The UI currently allows you to create variable with the same name (even with matching case).

jthrilly: Yes, sorry. Major doh from me on this one. Just trying to hurry it through! Will fix now.

PR #587, opened 2019-10-07
Release branch for 4.2.0

jthrilly:
This release includes:

- Schema version 2
- Stage level node filtering
- Skip logic including ordinal and categorical vars
- Case sensitive variable creation/deletion
- Codebook that allows variable deletion


jthrilly:
Needs latest NC (supporting schema 2) to be pulled in before testing and release.

PR #591, opened 2019-10-16
Add missing paths to variable index & fix stage name formatting in codebook

wwqrd:
- Add missing variables to variable indexes
- Fix stage name formatting in codebook
- Use development protocol to test against usage
- Update assetIndex to omit 'text' content.

Resolves: https://github.com/codaco/Architect/issues/590


wwqrd:
Oh thanks, I'll make sure to remove the duplicates!

wwqrd:
> The scrollbars for the "used in" column are not super attractive, but everything works well and looks correct!

Agreed, it was a quick fix to make the list readable. I think a tooltip might be a better fit.

PR #599, opened 2019-11-19
Update nc submodule reference

wwqrd:
Update NC ref to include: https://github.com/codaco/Network-Canvas/pull/997

Will need to be updated once that is merged.


rebeccamadsen:
Updating the nc submodule does take care of the skip logic error.

PR #605, opened 2020-01-02
New input types

wwqrd:
UI for configuring new input types

Requires:
https://github.com/codaco/protocol-validation/pull/28
and https://github.com/codaco/Network-Canvas-UI/pull/116


jthrilly:
I've noticed a couple of issues with this:

- The anchor date of the relative date picker seems to have no effect.
- Changing the variable from one datetime variable to another does not correctly update/reset the rest of the edit field form

src/components/Parameters/Parameters.js
```diff
@@ -22,7 +22,7 @@ const getComponent = (options) => {
 
 const Parameters = ({ type, component, ...rest }) => {
   const ParameterComponent = getComponent({ type, component });
-
+  console.log('parameters', rest);
```
wwqrd: ♻️

jthrilly:
Okay, all changes above have been implemented. These are:

- Updated min, max, and anchor date inputs to use DatePicker component for selecting a date.
- Implemented a toggle to "use the interview date", which defaults to on. When deselcted it shows the anchor date input selector, which is required. When the toggle is changed back to on, the value of the anchor date is reset to null.
- Updated UI to bring in fixes for sctollTo and reset.
- Updated the wording throughout
- Updated the "date resolution" select box in DatePicker to automatically reset the min and max inputs so that they can be set correctly. Could have updated them to reformat their value to match the new requirements, but thought specific reset was better.

wwqrd:
Okay, this looks good to me, though I can't review it since it's my own PR 😉

PR #613, opened 2020-01-28
Add support for other option for the CategoricalBin interface 

wwqrd:
Adds a toggle option for categorical interfaces allowing the addition of an 'other' variable. This is a new text variable separate from the original categorical variable, with a custom label.

Also includes refactoring of the NewVariableWindow to better allow multiple instances.


wwqrd:
@jthrilly Looks good to me, I can't review this one since it's my own PR. @rebeccamadsen?

PR #620, opened 2020-04-01
Fix inline variable delete (add where missing and remove where variable is in use)

### LINKED ISSUE #572: Don't show the delete cross next to variables that cannot be deleted
In the following screenshot, I used `discussFreq` elsewhere in the interview, so it cannot be deleted from this select dropdown. Clicking the 'X' does nothing.

Screen Shot 2019-08-23 at 11 07 21 AM

I think it would be better to:

1. Not show the cross when a variable can't be deleted
2. Show a small pill to let the user know a variable is unused (a pattern we use elsewhere, that we should make universal to showing a variable
3. Prompt to confirm deletion. Unsure about this one - perhaps not needed given the variable is unused.

Something like below:

Screen Shot 2019-08-23 at 11 12 48 AM

wwqrd:
This introduces a new method of checking for variable usage by using JSON.stringify and assessing the protocol state (and form state) as a string.

It is more robust than the indexer in the sense that no prior knowledge of protocol shape is needed to assess usage.

See notes here for alternate approaches: https://github.com/codaco/Architect/issues/572#issuecomment-607424246

Resolves  #572


rebeccamadsen:
Which screens should I see this delete option on? I can see and use it on Name Generator, for example, but can see and not use the delete on Ordinal Bin. Just wanted to check where I should be expecting this to work.

jthrilly:
> Which screens should I see this delete option on? I can see and use it on Name Generator, for example, but can see and not use the delete on Ordinal Bin. Just wanted to check where I should be expecting this to work.

Ideally, since the parent issue was about consistency of UI, this would happen anywhere there is a variable chooser. I think right now it might only be in the form field editor. 

@wwqrd can we expand it to the variable chooser more generally? Do you think that should be a separate PR? Think I know the answer ;)

wwqrd:
I'll take another look and check it's consistently in all pickers.

wwqrd:
Based on the code, these may need access to the new functionality, will review them next.

- QuickAdd
- NarrativePresets
- NodeType
- AssignAttributes
- SociogramPrompts

wwqrd:
[x] - QuickAdd
[x] - NarrativePresets (should group variable also have it - it's just a regular select?)
[x] - NodeType (changed to a simpler selector as doesn't need this info)
[x] - AssignAttributes (left existing selector, but don't think delete functionality makes sense here?)
[x ] - SociogramPrompts

wwqrd:
@rebeccamadsen to summarise, the interfaces that use this functionality (and might want reviewing) are:

- Categorical Bin > Prompts
- Ordinal Bin > Prompts
- Narrative > Narrative Presets
- QuickAdd Name Generator > Quick add Variable
- Sociogram > Prompts (highlights/layout)
- NameGenerator > Form > Field (use a different backend because this doesn't actually doesn't create variable until field is saved)

Omitted is the Sociogram Edge CreatableSelect, that would be deleting edges rather than variables - should that fall under the same PR? @jthrilly

rebeccamadsen:
Categorical bin, NameGenerator, Sociogram, and Quick add variable all work for me!

Narrative presets doesn't show the "x" on group variables, but I can't create those either, so maybe that makes sense; I can delete unused layout variables on Narrative.

Ordinal Bin, I can see the "x" next to unused variables, but pressing it doesn't do anything. Let me know if you need more info.

wwqrd:
Agreed on the group variables, I wasn't sure either. I think the main reasoning for this feature is so that you can delete accidential mistakes when creating new variables inline, so I left it out. Perhaps you should be able to create group variables here?

I will look into the Ordinal bin.

rebeccamadsen:
I could understand both directions for Narrative. Why would you be creating variables in Narrative? That's not what it's for. You could end up with variables being used there that aren't anywhere else, not exactly the purpose of Narrative. But if the uniform experience trumps the logic of trying to keep the user from doing something silly...?

wwqrd:
I fixed the ordinal bin although I've run into another hurdle.

The new variable usage feature allows us to check for usage in the protocol and in the currently editing forms. There may be false positives, because the state is duplicated several times - the protocol, a form containing the stage, a form containing a single prompt of that stage; this can then prevent the option component from showing the delete button. **This is compounded by the fact that the delete variable redux action refers strictly to the protocol when checking for usage**.

Here's the problematic scenario I ran into today:
1. User edits a prompt, and creates a new variable
2. Selector thinks both are in use because whilst the form doesn't contain the old variable, the protocol state does. So no delete button is shown.

I figured we could update the selector to allow selective removal of paths, e.g. the currently edited stage (in implementation we _also_ have to delete the variable from one of the forms too, since we are editing a single prompt in _another_ form):
1. User edits a prompt, and creates a new variable
2. Selector now correctly shows the variable as not in use and delete button is visible
3. Delete action doesn't do anything, because the action checks against usage in the protocol only (which hasn't changed yet because we haven't saved).

All this stems from the fact that we have multiple disagreeing states. Which I attempted to resolve in the "live" protocol editing branch which I thought was too big a change for this feature, but I'm starting to wonder.

A quick fix from this point might be to blindly trust the usage state from the prompt, but if we do that and then cancel the prompt changes we now have a stage without a variable.

jthrilly:
> I could understand both directions for Narrative. Why would you be creating variables in Narrative? That's not what it's for.

This is correct. The narrative is purely for displaying existing variables. Perhaps this should be updated to be a simple "chooser", assuming we still have one of those.


@wwqrd on your issues:

> * User edits a prompt, and creates a new variable
> 
> * Selector thinks both are in use because whilst the form doesn't contain the old variable, the protocol state does. So no delete button is shown.

I think this is the correct behaviour. The old variable can't be said to be unused until the outcome of editing the prompt is known. Similarly, the new variable can't be said to be used/unused/existing until the outcome of that.

wwqrd:
Glad to hear it! 

I've reverted the behaviour to the original implementation.

rebeccamadsen:
Works for me, once the linting does!

PR #622, opened 2020-04-08
Fix crashes when importing assets

### LINKED ISSUE #615: Architect crashes when a larger audio file added to asset library
When I tried to add an MP3 audio file of 28.3 MB to the asset library, the application kept crashing every time I tried to add it. I had the audio file saved in my downloads folder, so would browse to add a file to the asset library, double click the file and shortly thereafter Architect would crash. This happened three times and every time I was prompted to reopen the application by a dialog that would open (see below). After three times with no success, I elected not to reopen per the dialog and instead just restart the application altogether. When I reopened Architect and tried again to add the file, I was able to without an issue. 

I grabbed the following details the second time the application crashed.
image (8)

wwqrd:
Architect was crashing intermittently when importing files (https://github.com/codaco/Architect/issues/615).

Stripped the implementation back to a plain `<input type="file" />`, and app was still crashing. 

Believe this to be due to the following reported issue in electron v5: https://github.com/electron/electron/issues/21025

As update is not feasible, this fix instead switches from an input to use uses Electron's `showOpenDialog` method (https://github.com/electron/electron/blob/5-0-x/docs/api/dialog.md), and native node `fs` to import the file.

Changes:
- Should be faster since we are now doing a `fs-extra` copy() rather than reading and writing the file in the JS render process.
- Included better error handling, distinguishing validation from import errors
- Added a 'loading' state to the asset handler.

Resolves #615

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3048,15 +3048,6 @@
             "integrity": "sha512-Wm6ukoaOGJi/73p/cl2GvLjTI5JM1k/O14isD73YML8StrH/7/lRFgmg8nICZgD3bZZvjwCGxtMOD3wWNAu8cg==",
             "dev": true
         },
-        "attr-accept": {
-            "version": "1.1.3",
-            "resolved": "https://registry.npmjs.org/attr-accept/-/attr-accept-1.1.3.tgz",
-            "integrity": "sha512-iT40nudw8zmCweivz6j58g+RT33I4KbaIvRUhjNmDwO2WmsQUxFEZZYZ5w3vXe5x5MX9D7mfvA/XaLOZYFR9EQ==",
-            "dev": true,
-            "requires": {
-                "core-js": "^2.5.0"
-            }
-        },
         "author-regex": {
             "version": "1.0.0",
             "resolved": "https://registry.npmjs.org/author-regex/-/author-regex-1.0.0.tgz",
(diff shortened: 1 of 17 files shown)


rebeccamadsen:
I am not sure if I can replicate the crashing on Windows, but I can look at how the changes work at least. I can open and drag files into the assets. However, some of the assets I used to be able to add I can no longer add. For example, opening the dialog to browse to an asset, I cannot see some of the files as an option (e.g. .png, .mov, .m4a, some .mp4 files). I can drag and drop the .mp4 file that wouldn't show in the dialog, but the .mov files won't allow that either. Does this change what files we accept?

wwqrd:
I've fixed a few bugs and made some cosmetic changes:

- File extensions with numbers (e.g.  mp3, mp4) were being excluded, this should now be fixed.
- Dropping files didn't give any user feedback as to why they were rejected, error messages are now displayed.

@rebeccamadsen I've added m4a as that wasn't included in the extensions list (before we used a mixture of mime-types and extensions, and now it's just extensions). Let me know if you think there are any more missing in the [extensions config](https://github.com/codaco/Architect/blob/47625f60f2abd64ce4a17b78af623196426b876d/src/config/index.js#L26,L31).

rebeccamadsen:
Hmm. How about .ogg, .wav, .bmp? I don't have any, but I think those would be supported? There's .heic files, but I don't think browsers support those.

I do have a few jpg files that have capital letters in the extension, like ".JPG", and those don't load here, but all the mov and mp4 files I have are just great!

The error message was a nice touch.

jthrilly:
> I do have a few jpg files that have capital letters in the extension, like ".JPG", and those don't load here, but all the mov and mp4 files I have are just great!

This sort of thing was why I wanted to use MIME type. The extension doesn't matter - the content is what matters.

Problem is that MIME is seriously broken! Particularly on Windows. I think the best we can do is flatten the case and hope we can maintain an adequate 'whitelist'.

> Hmm. How about .ogg, .wav, .bmp? I don't have any, but I think those would be supported?

I think in general we should be okay to limit people to "sensible" formats. The main criteria would be size as we don't want to encourage large protocols. I think that means BMP and WAV and out. @wwqrd, I will let you decide specifically. One thing is that they **must** be not only cross browser, but cross platform compatible. This particularly matters for video as there's a bit of an ongoing codec war between Apple and Google.

wwqrd:
- Added capitalization support, and `mp4` for audio, but I'm agreed on the uncompressed formats and from the docs it doesn't look like ios safari supports `ogg`.
- Added some info to the file not support error message which lists the supported extensions.
- Fixed a bug where the uploader could not be used after the user clicked to open a file and then cancelled.

Think this is ready now?

PR #625, opened 2020-04-20
Improved Codebook

wwqrd:
- Adds Ego section to the codebook, so that ego variables can be deleted
- Shows delete entity button no matter what the usage status is - and instead notifies user if the entity cannot be deleted.
- Makes "used in" clickable (takes you to edit stage)

Known issues:
- Animation to edit screen isn't great, planning to address that as part of: https://github.com/codaco/Architect/issues/616

Resolves https://github.com/codaco/Architect/issues/619
Resolves https://github.com/codaco/Architect/issues/603
Resolves https://github.com/codaco/Architect/issues/618


jthrilly:
Does this also make the "used in" items clickable (to take you to that stage)?

PR #628, opened 2020-04-28
Improved loading and saving ux

### LINKED ISSUE #616: Improve user experience for asynchronous file system actions
On slower systems async actions show no indication that they are working, which is confusing and can lead to strange behaviour if actions are repeated.

For the following features:

- [x] Loading an asset
- [ ] Saving a protocol
- [ ] Opening a protocol

Async actions should be updated to:

1. Indicate progress (by using a spinner for example)
2. The UI should block the same action being called repeatedly

Potentially related: https://github.com/codaco/Architect/issues/615

wwqrd:
Changes to improve the resiliance and user experience of saving and loading protocols, as well as some refactoring to streamline some of the processes.

- All open/save dialogs are now triggered in the renderer window. Previously they were a mixture dependant on where they were used (menu vs. in-app). Having a single process for both means the they remain consistent
- Adds saving animation to the protocol save button, which is also deactivated during save
- Adds a loading screen animation for larger protocols (triggered after half a second).
- Introduces the concept of "locks", which can be named and stacked, this allows us to prevent multiple actions being dispatched of the same type, or even accross action types (if they use the same lock name).
- Collates the in app-dialogs to delegate to a single set of "promisified" methods (that mirror later Electron versions)
- In app dialogs open attached to window so focus can't be lost (macOS)

Known issues:
- ~When a protocol is already open, previous protocol should probably be cleared immediately from view.~ Updated openProtocol and the unbundleAndLoad thunks to reset the session. Also added a change which resets the preview.

Resolves #616


rebeccamadsen:
I can't seem to use "Open Existing" from the start screen, or the "File->Open..." when on the start screen. I get this error:

```
TypeError: Cannot read property 'filePath' of undefined
    at http://localhost:3003/static/js/bundle.js:227924:44
```

Works great if I open a previously opened protocol, and I can use "File->Open" from within a protocol.

wwqrd:
@rebeccamadsen That issue should be resolved now!

PR #629, opened 2020-05-05
Skip logic in stage editor

### LINKED ISSUE #517: add ability to remove skip logic
If a user changes their mind about having skip logic on a stage, we should have some way to remove it without deleting the whole stage.

Once deleting skip logic is possible, it would make sense to also require existing skip logic have at least one rule, and to require "Any" or "All" to be selected.

I wonder about whether it makes sense to have an indicator either in the timeline or the stage view that a stage has skip logic on it? Right now the only indication is to click on "configure skip logic" again or to drill into the code view.

### LINKED ISSUE #617: [META] Improve stage level node filtering and skip logic location
We have had feedback that Skip logic and stage level node filtering are confusing concepts for users. 

This is not helped by the fact that they have similar names, but are located in very different parts of Architect.

We therefore want to move these sections to the stage editor.

Key features:

- [ ] Will resolve https://github.com/codaco/Architect/issues/517 (ability to remove skip logic from a stage)
- [ ] Will add **two** expandable sections ("Skip Logic" and "Network Filtering") that be default are collapsed so it does not take up significant space. If either skip logic or node filtering are enabled for a stage, the sections should be expanded in the editor view when opened.
- [ ] There should be an icon on any stage in the timeline view when it uses either feature (separate icon for each).

### LINKED ISSUE #632: Ordinal variables aren't working with skip logic
There is a bug preventing the use of ordinal variables in skip logic. When a rule is being defined the value section is greyed out and cannot be modified.

wwqrd:
- Update the skip logic to be edited in the stage editor rather than on a separate screen.
- Update the filter section to be stylistically similar to skip logic panel
- Integrate node/edge filtering with the node/edge selector.
- Update StageEditor with improved header section
- Strips out unused guidance sections.
- Fixes broken ego ordinal and categorical rules.

Known issues:
- Unsure about context panel color choices: grey when on white backdrop and purple when on grey backdrop (see node panels on a name generator screen)

resolves #632
resolves #617
resolves #517


rebeccamadsen:
Is this ready to review? I like the header and the unifying of styles.

When I uncheck a filter or skip logic, it prompts me that I will clear the filter, but if I cancel this dialog instead of choosing to clear filter, I still lose the rules. This isn't a problem with the node panel filter.

Can there be some white space on the left in the new node type screen?

Preview was acting strangely for me, but I'm not sure if that's related. Clearing or not clearing a filter didn't always have an effect. I'll try to see if I can get more specific about this tomorrow.

wwqrd:
> When I uncheck a filter or skip logic, it prompts me that I will clear the filter, but if I cancel this dialog instead of choosing to clear filter, I still lose the rules. This isn't a problem with the node panel filter.

Fixed this, looks like the name gen panels had the opposite problem - they were never cleared!

> Can there be some white space on the left in the new node type screen?

Nice spot this wasn't mean to have changed, will look into it.

wwqrd:
> Can there be some white space on the left in the new node type screen?

This is a knock on effect of some of the styles I changed for the stage editor screen. I've shimmed a fix of sorts by manually adding some `editor` class divs, but it's all kinds of not pretty. Acceptable for now?

rebeccamadsen:
How is a filter deleted? For example, if I have no filter on an OrdinalBin nodetype, and then create one, the preview acts as I expect. Then if I clear the filter, it's as if every node is filtered out of my OrdinalBin. I would expect no filter/cleared filter meaning nothing is filtered from my view and I should be able to see all the nodes?

Similarly with the node gen side panels. Clearing a previous filter means that all the nodes are filtered out.

Is there a way to delete a node side panel once you've started making one? I don't think that's related to this, I was just curious about it when experimenting.

Generally looking good though! Filters otherwise work as I expect, and I checked the ego cat/ord variable skip logic as well. Spacing looks good visually; shims for the interim work for me!

PR #639, opened 2020-05-26
Dyad Census Interface Editor

wwqrd:
Adds support for editing the Dyad Census interface type.


rebeccamadsen:
How do you create a new dyad census stage? I didn't see an option for it.

wwqrd:
Okay, that should be in now. As well as a preview for the interface type.

Have added a new issue to update the other preview images now that NC has changed. https://github.com/codaco/Architect/issues/643

Issue #623, opened 2020-04-10
Creating a variable with the same name (but different type) throws an error in console but does not notify the user


wwqrd:
The reducer throws an error and doesn't provide feedback.

Steps to reproduce:
- Create a sociogram interface with new layout variable
- Create a quickadd interface, and attempt to create a new variable with the same name.

PR #646, opened 2020-06-18
Warn about duplicate variables in creatable select

wwqrd:
When a variable is created inline the variable name was only compared against those of the same type, but we require variable names to be distinct accross types.

This adds functionality to supply an additional 'reserved' list to CreatableSelect, and a new component VariableSelect which automatically populates this list based on the entity type.

Some of the improvements seen here could be taken further, but due to the impending release I've left it here.

Possible improvements:
- Do variableOptions and reserved lists both need to exist for the validation check?
- Could we also have specialised components for other CreatbleSelects, e.g. edge type selector?
- Since we pass `entity` and `type` to the variable select it's no longer necessary to pass the createVariable handlers to the parent component, it could be completely handled by VariableSelect to simplify PromptFields

Resolves: #623

### DIFF
--- a/src/components/Form/Fields/CreatableSelect.js
+++ b/src/components/Form/Fields/CreatableSelect.js
@@ -19,6 +19,12 @@ const getValue = (options, value) => {
 };
 
 class CreatableSelect extends PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = { reservedVariable: null };
+  }
+
   get value() {
     return getValue(this.props.options, this.props.input.value);
   }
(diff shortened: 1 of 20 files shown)


rebeccamadsen:
This compares the variable within the node type, not across all node types, right? So different nodes can have "name" fields?

Not sure if it's related, but I tried adding a variable to the quick add name generator, and it shows up in the codebook as "not in use."

wwqrd:
> This compares the variable within the node type, not across all node types, right? So different nodes can have "name" fields?

Yes that's correct, only within the entity. It's intended to prevent duplicate names where they might not show up in the options list as an existing variable. e.g. a field variable input (which won't show layout variables), and a sociogram a layout input (which won't show text variables).

> Not sure if it's related, but I tried adding a variable to the quick add name generator, and it shows up in the codebook as "not in use."

I think this is unrelated, it shows up on master too. Will add an issue for it!

wwqrd:
@rebeccamadsen https://github.com/codaco/Architect/pull/653 This should fix the codebook issue.

PR #651, opened 2020-06-22
Constrain variable names

wwqrd:
Variable names must use only letters, numbers, the symbols ._-: and contain no spaces.

This updates UI to accomodate the rules, as well as adding validation to external data imports.

Note: This builds off of https://github.com/codaco/Architect/pull/646 which will need to be merged first.

Resolves https://github.com/codaco/Architect/issues/647
Resolves https://github.com/codaco/Architect/issues/645


wwqrd:
Force pushed a rebase that fixed merge conflict.

rebeccamadsen:
The auto migration only happens with a previous schema version, not the current schema, right? Just making sure that's expected. I had a schema 4 protocol that just gave me error messages about variable names, but older schema protocols went through the migration just fine.

The width of the options label and value fields changes when typing invalid characters. It's a little disconcerting:
options 2
options 1

Are edge types allowed to have the invalid characters? I was able to create edge types within sociogram that had spaces and backslashes and things. Not positive if that is a problem. But it's at least inconsistent because creating node types and edge types elsewhere don't allow those characters:
edge creation

jthrilly:
> Are edge types allowed to have the invalid characters

Just on this, I think the only requirement for entity names is that they be unique, although we do use them for file names (behind a sanitize function). I think allowing these characters is fine, but blocking them (if it isn't too much work) would probably be neater.

wwqrd:
> The auto migration only happens with a previous schema version, not the current schema, right? Just making sure that's expected. I had a schema 4 protocol that just gave me error messages about variable names, but older schema protocols went through the migration just fine.

The migration is from 3 -> 4 we've been using version 4 in development for a while so this is confusing - but there hasn't been a public release that uses 4 as far as I know.

> Are edge types allowed to have the invalid characters? I was able to create edge types within sociogram that had spaces and backslashes and things. Not positive if that is a problem. But it's at least inconsistent because creating node types and edge types elsewhere don't allow those characters:

Agree with you on this, not sure if I just missed this one, or edges entirely but I will go back and check.

> The width of the options label and value fields changes when typing invalid characters. It's a little disconcerting

I think this is unrelated to this PR. If this is a quick fix I'll resolve it, but this component isn't that well designed so it might not stick around for much longer.

PR #659, opened 2020-07-08
Update to latest node/electron (match Network Canvas)

wwqrd:
Updates dependencies to match those used in Network Canvas.

- node, 12.8.0 -> 12.14.1
- npm, 6.10.2 -> 6.13.4
- electron, 5.0.9 -> 9.0.0 😃 

I haven't been able to test builds on my system, maybe a key signing thing? But otherwise this seems to work fine.


wwqrd:
> I could drop an asset file, but using the "click to browse" dialog gave me an error:
> 
> `Uncaught TypeError: Object(...)(...).then is not a function`
> 
> I also get an "open error" if I try to open a protocol from the File->Open dialog. Similarly with the File->Save a Copy.

Hmm, I have a feeling this may have been fixed by https://github.com/codaco/Architect/pull/659/commits/9a2a796bb6f22426d0388c438069e61f31fd3d8a? I'm not seeing the same errors on my end?

rebeccamadsen:
With the latest, I get this, plus a white screen:

```
EditorScreen.js:99 Uncaught TypeError: Cannot read property 'bool' of undefined
    at Module../src/components/Screen/EditorScreen.js (EditorScreen.js:99)
    at __webpack_require__ (bootstrap:789)
```

I tried blasting away my node_modules in case something was out of sync, but couldn't get past this.

wwqrd:
Closing until this is ready to review.

PR #673, opened 2020-10-27
Fix support for additional attribute variable types

### LINKED ISSUE #670: Assign additional attributes doesnt work with new variables of type datetime or scalar
As per title: creating a new variable of type scalar or date results in an empty variable when returning to the assign additional attribute screen:

Screen Shot 2020-10-08 at 3 45 12 PM

wwqrd:
- Support scalar and datetime in assign attributes
- Add option for create new selector without inline support

Fixes #670

### DIFF
--- a/src/components/AssignAttributes/Attribute.js
+++ b/src/components/AssignAttributes/Attribute.js
@@ -40,7 +40,7 @@ const Attribute = ({
           label="Variable:"
           component={NativeSelect}
           options={variableOptions}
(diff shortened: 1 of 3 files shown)


jthrilly (approved):
Looks great. Thanks for the quick fix!

jthrilly:
It looks like travis won't run because this branch targets the refreshed home branch. If the tests pass locally, please go ahead and merge it.

PR #674, opened 2020-10-27
Show error to user when bundle cannot be saved

### LINKED ISSUE #671: Bundle errors should alert the user
A user encountered an error while saving that provided no user feedback:

Screen Shot 2020-10-13 at 2 50 41 PM

In this particular case, the EPERM should be a specifically handled error message that tells the user that the file could not be written due to a permissions error, and tells them what they can do. What they can do is (1) Check that the file is not "read only", (2) Delete or rename the file and see if architect can recreate it

wwqrd:
- Ensure bundle protocol rethrows error after dispatching failed action
- Add error message for EACCES error

Fixes #671


jthrilly (approved):
Nice! Looks great!

PR #691, opened 2021-01-11
Organise netcanvas file for better testing

### LINKED ISSUE #689: Migration to V4 protocol version fail
Hi,

When I try to migrate from v3 to v4, I have this error :
Screenshot from 2020-12-24 00-50-16
I'm on Ubuntu 20.10, and my /home is on another partition than /tmp.

I asked to write the new version to /tmp, and it worked. But I have the same problem when I copy the protocol to my /home, update it, and try to save it.

As a workaround, I work on another partition for the moment.

Thank you for your work.

Laurent

wwqrd:
Separate out interdependent functions in netcanvasFile so that module can be more thoroughly tested.

Also resolves #689


rebeccamadsen (commented):
Looks fine for me on windows, but are these specific things you want tested on specific devices?

wwqrd:
Nothing specific, it's mainly just a refactor, functionality shouldn't change apart from the rename/copy fix, which I verified on Linux. Issue may apply to Windows too, to test you could try saving to a usb stick or secondary partition.

jthrilly (approved):

PR #702, opened 2021-03-11
NPM 7 upgrade

jthrilly:
- Updates to NPM7
- Removes several dozen unused packages, and updates many others
- Changes eslint and babel configuration


.eslintrc.json
```diff
@@ -22,8 +22,10 @@
     "rules": {
       "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }],
       "react/forbid-prop-types": "off",
+      "react/jsx-props-no-spreading": "off",
```
jthrilly, line 25: I think I know what you'll say, but I don't like this rule being off in a blanket sense. The other two apps are able to keep it on, and only disable it for tests or other specific and intentional instances. In general I feel like the reasons for this rule are 'good', in that it prevents opaqueness surrounding which props a component is getting, and I do think this has been a source of bugs.

As I said, I think you'll probably argue that Architect makes far more egregious use of the HOC/recompose patterns, and that it will be much harder to untangle props because of this. If that's the case I'm happy to leave it off - just wanted to check in about it, really.

wwqrd: We can always add exceptions inline. I think I prefer doing it that way generally as it forces you to consider if it's really necessary.

wwqrd:
I've noticed some bugs with this which I'm looking into now

wwqrd:
Superseded by https://github.com/complexdatacollective/Architect/pull/708

PR #709, opened 2021-04-09
Update to use new RichText field for markdown content

wwqrd:
- Replaces TextArea with RichText where markdown is useable.


jthrilly:
I spotted a couple that are missing:

- Form field previews don't render the field label markdown. This might require rethinking the styling a bit, as we also have the input type/component badge in the preview component.
- Ordinal and categorical labels are now also inline markdown
- Followup question for categorical bin: both the label for bin and the followup prompt should be inline markdown

Otherwise looking great 👍

wwqrd:
- [x] Form field previews don't render the field label markdown. This might require rethinking the styling a bit, as we also have the input type/component badge in the preview component.
- [x] Ordinal and categorical labels are now also inline markdown
- [ ] Followup question for categorical bin: both the label for bin and the followup prompt should be inline markdown

jthrilly (approved):
So hot.

PR #714, opened 2021-04-15
Release branch for 6.1.0

jthrilly:
## Network Canvas Architect 6.1.0

This is the first significant feature update to Architect since our launch. These changes required **a new protocol schema version**, which means that any current protocols opened in this version of Architect will be automatically migrated to the latest schema version. Review the changelog below, and take extra care when updating if you are currently in the process of collecting data.

### Changelog:

- **Added a new interface: the "Tie-Strength Census"**. This interface follows the same paradigm as our dyad census, except it assigns a value to an ordinal variable if an edge is present. This can be used to capture both the strength and existence of a relationship in one interaction. It resulted from a collaboration with Dr. Brea Perry at the Indiana University Network Science Institute. 
- **Introduced a new input component: "BooleanChoice"**. This new component is available for boolean variable types, and differs from the existing toggle switch in two ways: (1) it requires user interaction to set a value (it has no default), and (2) it supports rich labels describing each 'choice'.
- **Introduced markdown support for field labels and option values**. This means that all field labels and option values for ordinal/categorical form variables can now contain limited rich text formatting, such as bold and italic text. A WYSIWYG editor has been added to make formatting these labels more straightforward.
- **Introduced new validation options: unique, differentFrom, and sameAs**.
  - **sameAs** - Requires that a variable value be the same as another variable value.
  - **differentFrom** - Requires that a variable value be different from another variable value.
  - **unique** - Requires that a variable value be unique among all entities of the same type.

# Focus for testing:

Please heavily test the new features! In particular, form interfaces and the new component, as well as the new interface! Look out for any visual oddities in existing components caused by changes to UI. Look out for situations where wider markdown support is creating issues.


rebeccamadsen:
- [ ] On Information stage, I can insert new lines in a text area, but <br> shows up in the Architect stage editor view (not in the Interviewer preview).
- [ ] On Information stage, I can use \~~ to create strikethroughs, and \``` and \` to create code, and \> for blockquote, and it looks good in the Architect view but is stripped from Interviewer.
- [ ] Narrative presets show markdown as active in the Architect preview (e.g. if I put asterisks around a word), but Interviewer doesn't allow it.
- [ ] NG Inputs from the Development protocol has new lines in a label, but goes funky with \<br>s all over if you edit it because labels don't allow new lines. Development protocol update?
- [ ] When selecting an ordinal variable in the Alter Edge Form or Alter Form as a field, that was previously created on either Tie-strength census or Ordinal Bin, all the previously created variable options show up. However, selecting a required input control removes all the options.
- [ ] If I answer the "decline" option on a Tie-strength census, should we already "know" the answer is "no" for a Dyad census too? We know the "yes" answers.

rebeccamadsen:
Funky \<br>s:
markdown-br

Development protocol with new lines in a label:
architect-label-new-lines

jthrilly:
>  On Information stage, I can use ~~ to create strikethroughs, and ``` and ` to create code, and \> for blockquote, and it looks good in the Architect view but is stripped from Interviewer.

I think blockquote is the only one to fix here. Strikethrough and code should be disallowed types, and out shortcut handler should be updated to detect them

PR #724, opened 2021-05-09
Printable protocol summary

wwqrd:
Adds the ability to preview and then print a summary of the open protocol.


jthrilly:
This seems to have a couple of missing features:

- [ ] Add a "Print protocol summary" button to the codebook view
- [ ] Restyle the preview window as a print preview (Electron doesn't support the Chrome print preview functionality)
- [ ] Make the filename reflect the protocol name

package.json
```diff
@@ -263,6 +265,7 @@
         }
     },
     "dependencies": {
-        "postcss": "^8.2.8"
+        "postcss": "^8.2.8",
+        "react-router-dom": "~5.2.0"
```
jthrilly, line 269: Should this be a dev dependency?

src/ducks/store.js
```diff
@@ -7,7 +7,7 @@ import ipc from './ipc';
 import { rootReducer } from './modules/root';
 
 const persistConfig = {
-  key: 'architect',
+  key: 'architect6.1',
```
jthrilly, line 10: Why was this changed? This will reset all app state for current users, won't it? We should revert unless it was intentional.

src/lib/ProtocolSummary/ProtocolSummary.scss
```diff
@@ -0,0 +1,3 @@
+.printable-summary {
```
jthrilly, line 1: Sorry about this - needs to be removed.

src/lib/ProtocolSummary/components/Stage/Presets.js
```diff
@@ -0,0 +1,77 @@
...
+  // eslint-disable-next-line react/forbid-prop-types
+  presets: PropTypes.object,
+};
+
+Presets.defaultProps = {
+  presets: null,
+};
+
+export default Presets;
+// edges: {display: Array(1)}
```
jthrilly, line 72: Remove

jthrilly (commented):
In addition to the issues I highlighted, we also need to fix the proptype console errors in the app.

src/ducks/store.js
wwqrd: Agreed, this shouldn't change unless the store has changed in some other drastic way (doesn't look like it has to me).

PR #725, opened 2021-05-10
Fix Radio field id

wwqrd:
Must use id.current with ref


wwqrd:
Superseded by https://github.com/complexdatacollective/Architect/pull/726

PR #731, opened 2021-08-06
Preview assets

### LINKED ISSUE #687: Ability to preview and/or export items from the resource library
Once data has been imported, it is very difficult for users to inspect it. We should add a simple "preview" component so that network assets in particular can be inspected. As a bonus, it would be great to be able to "save" assets back out of the protocol and back into user space.

wwqrd:
Add preview functionality to asset browser, this is particularly useful for network data.

Assets can also be downloaded for viewing outside of the app.

Resolves #687

controls

preview


rebeccamadsen:
I super love this feature, let me just say. I ran into no major issues, and only a couple of small things.

On loading json preview in development protocol, I got a warning:
```
Warning: Each child in a list should have a unique "key" prop.
```
I kind of think this is an artifact of the oddness of the json for the development protocol, because it on purpose doesn't have the "name" variable correct, so some show up as null?

Also really minor, but the development quadrant images don't show up on a white background. I'm not really recommending anything, because they download just fine, but I suppose an above and beyond solution could be to offer to toggle the background color for images, however I'm not convinced this is necessary.

wwqrd:
Thanks @rebeccamadsen I've resolved those warnings.

When looking at this afresh I thought it also make sense to add the preview/download functionality to the Codebook.

rebeccamadsen (approved):
Adding it to codebook was a great addition. Looks good to me!

Issue #677, opened 2020-10-29
[META] New Stage Screen Improvements


jthrilly:
The new stage screen would benefit from a revision.

- [x] Interfaces should be displayed in a filterable list
- [x] Each interface section should show the interface name, a its preview image, and perhaps also text about it.
- [x] In additional to the list being filterable by name, it should also be filterable by "tags"
- [x] Tags should be interface affordances/functionality: "Create nodes", "Create edges", "Ego data", "Add Node Attribtues", "Add Edge Attribtues", "Use Roster Data"
- [ ] Revisit the idea of a "wizard" that can reccomend an interface. Could be a simple branching tree. Ask the user: "Do you want to create an entity, add attributes to entities, or both?" and so on.

PR #733, opened 2021-09-08
Add rename variable feature

wwqrd:
Updates codebook and inline variable selection to include a rename variable dialog.

Delete button in codebook has been changed to an icon to create more room for the rename button.

**Rename dialog**
dialog

**Codebook view**
codebook_rename

**Inline view**
inline_rename

Resolves #732

PR #734, opened 2021-09-23
 Improved new stage screen

wwqrd:
Updates New Stage Screen to a filterable list. This is seen when the user inserts a new stage into a protocol.

### Features

- [x] Stages can be searched by name/keywords
- [x] Stages should be filterable by tags
- [x] Stages should show a description

**Main screen**
main-screen

**Search results view**
results

Relates to #https://github.com/complexdatacollective/Architect/issues/677


jthrilly:
UI Feedback

- [ ] I think this should be a modal as opposed to a full screen view
- [ ] Needs a heading area and short explanatory text
- [ ] Side panel should be fixed width and pinned to the left edge, with the rest of the space taken up by a responsive grid of items. Grid columns should change with viewport width.
- [ ] Grid card style needs to be revised. I suggest trying heading at the top, then image, then tags, with a background for the card itself
- [ ] Categories and tags are mutually exclusive. We should just use tags.
- [ ] Tags should be cumulative, so that all must match rather than any in order for an interface to be shown.
- [ ] Sidebar tags should be colored all the time, and use a different style (opacity? outline?) for when they are not in use
- [ ] Sidebar tags should have more padding
- [ ] Tags should use the all caps and increased letter spacing style 
- [ ] Main input should be a filter, not a search. Filter should be applied after any filtering done by tags being selected.
  - [ ] Filter should either be part of the list component, or should be at the top of the side panel
  - [ ] Animations don't seem right (the separation of the exit animation and the layout animation is jarring). Perhaps this is a candidate for the hyper list in UI that I am working on?

wwqrd:
I've implemented most of the changes in a basic way, with some exceptions:

- Rudimentary keyboard navigation has been implemented using `tabIndex`, and a `onKeyUp` handler.
- Dialog window has not been implemented. There is a suitable component in the Rename Variable PR (https://github.com/complexdatacollective/Architect/pull/733), it would make sense to use this component once that is merged. 
- Related to the above, responsive grid layout changes have not been implement, as a `<Screen />` component this mirrors the layout of all screens in Architect with a fixed width of `80rem`.
- Some of the changes probably would make sense to implement once the `<HyperList />` component is ready including: list animations, grid layout changes, full keyboard navigation.

jthrilly:
I'm going to complete this work in #741

PR #740, opened 2021-12-10
Automatic node layout on the Sociogram

wwqrd:
Adds support for automatic layout on Sociogram: https://github.com/complexdatacollective/Interviewer/pull/1156

Screenshot 2021-12-10 at 12 29 23


jthrilly (commented):
Obviously this one is still pending feedback from the others a bit, but one thing I thought from first seeing it is that this feels more like an "on/off" type feature.

In other words, if it is enabled, the automatic layout runs immediately upon moving to the sociogram stage. This would make the toggle box a more appropriate control. What do you reckon?

wwqrd:
> In other words, if it is enabled, the automatic layout runs immediately upon moving to the sociogram stage. This would make the toggle box a more appropriate control. What do you reckon?

Are you saying that Sociogram would always have the automatic layout available, and this option only toggles the initial state? In that case - yes I think it would make sense.

As the feature currently works (force simulation is either available or not) I think this UI affords a bit more clarity on what is happening (like the fact that you can still manually position nodes if you want to). Maybe we could achieve the same in the intro text?

jthrilly:
I think the feedback above has been superseded by the Slack discussion. We settled on a play/pause button for this.

rebeccamadsen (approved):

PR #747, opened 2022-02-03
Fix/validation initial state

rebeccamadsen:
Ensures validation is reset to `null` so that it doesn't start off enabled when there are no existing validations.


jthrilly (approved):
Nice one :) Welcome to the world of 'weird things Architect does'

PR #758, opened 2022-03-17
Feature/min max namegenerators

jthrilly:
- Implements the configuration of min/max alter counts on name generator interfaces
- uses schema 7
- Includes changes to validations:
  - Ported over some implementations from NC
  - Ported over tests from NC
  - Fixed bug related to minValue and 0 validation


rebeccamadsen:
Looks great, other than missing an error message:

image

rebeccamadsen:
Pushed a fix for the error message (and the options error message as well). I'm good with this if you are.

rebeccamadsen (approved):

PR #763, opened 2022-03-30
protocol summary variable pills and behaviors

rebeccamadsen:
This fixes the existing variable pills in the protocol summary (for skip logic and filters), which were not getting the correct types and colors.

This also fixes the behaviors in the protocol summary to include the latest behaviors, and only to show behaviors relevant to the existing stage. It assumes relevance if the behavior is present.

Finally, this adds "pills" for all remaining variables in the protocol summary.

Screenshots from the development protocol:
image

Another:
image

A third:
image

Rinse:
image

Repeat:
image

### DIFF
--- a/src/lib/ProtocolSummary/components/Rule.js
+++ b/src/lib/ProtocolSummary/components/Rule.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { compose } from 'recompose';
+import RuleText from '@components/Query/Rules/PreviewText';
+import withDisplayOptions from '@components/Query/Rules/withDisplayOptions';
+
+const Rule = ({
+  type, options,
+}) => (
+  <RuleText type={type} options={options} />
+);
+
+Rule.propTypes = {
+  type: PropTypes.string.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  options: PropTypes.object.isRequired,
+};
+
+export default compose(
+  withDisplayOptions,
+)(Rule);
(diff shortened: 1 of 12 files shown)


jthrilly:
Looks really great! Much much better.

Only thing I see is that I think the rule section needs a little bit more work.

- The rule container should be full width
- The entity icon should be smaller
- Ideally, the styling should match how it appears in architect more closely

Perhaps beyond the scope of this, but it would also be nice to use the new edge entity icon. What do you think?

rebeccamadsen:
How about this?
image

And:
image

I think edge entity icons are already being used? Or is there something newer?
image

jthrilly:
The other place is in the subject section, but it looks like the node glyph isn't used there either, so lets just keep things as they are.

src/lib/ProtocolSummary/components/Rule.js
```diff
@@ -0,0 +1,21 @@
...
+const Rule = ({
+  type, options,
+}) => (
+  <RuleText type={type} options={options} />
+);
+
+Rule.propTypes = {
+  type: PropTypes.string.isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  options: PropTypes.object.isRequired,
```
jthrilly, line 16: Just a small thing with this, but if you are using the github co-pilot beta, it is quite good at auto-completing these.

jthrilly (approved):

src/lib/ProtocolSummary/components/Rule.js
rebeccamadsen: will look into it. :)

PR #765, opened 2022-04-04
fix asset preview for item

rebeccamadsen:
Images, videos, and audio were not showing correctly in the Stage. This restores the correct import so that the items will display correctly.


jthrilly (approved):
Great find

PR #773, opened 2022-08-16
Feature/install sample protocol

xchen0326:
This adds a installing sample protocol for demonstration from online feature. By clicking on the button in the welcome header card, users can download the sample protocol and automatically open it.


src/utils/fileSystem.js
```diff
@@ -0,0 +1,696 @@
+/* eslint-disable global-require */
```
jthrilly, line 1: You need to think more critically about if you need to bring over files like this. In this case, you definitely do not!

The only method from this file that you are actually using is `writeFile`, which as you can see below is a simply a wrapper around `fs.writeFile`. Why didn't you use this directly? `fs` is a core node module, and is included in Electron automatically. 

By importing this whole file, you duplicated many existing filesystem methods that Architect already has, which creates confusion and introduces issues with maintainability. Another issue is that this file is very large, and all the unused code you imported is simply wasted space.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+ * If the URL points to an instance of a Network Canvas Server, then the caller must ensure
+ * that the SSL certificate has been trusted. See {@link ApiClient#addTrustedCert}.
+ *
+ * @param {string} uri
+ * @return {string} output filepath
+ */
+const downloadProtocolFromURI = inEnvironment((environment)=> {
+
+  if (environment === environments.ELECTRON) {
+    const request = require('request-promise-native');
```
jthrilly, line 86: This is an npm dependency that you did not add to the `package.json` of the project. The only reason this code works is because other dependencies import this module. This is very bad for two reasons:

1. You do not determine the specific version of the dependency that is being imported. This means that you do not know the specific API for using the module.
2. Upgrading this dependency is out of your control. If a transient dependency is upgraded, this module version will likely change. It may even be removed entirely, which will cause the software to crash or not build.

All dependencies must be specified in `package.json`!

However, there is a more significant issue here. Remember that I mentioned that there is a great deal of legacy code in the project? Here, the module you have used is marked as no longer supported by its author: https://github.com/request/request-promise-native#deprecated

This means that we should not introduce any additional uses of this module anywhere in our code, and we should refactor any code that is using it when we change it. 

You should think more critically about what a dependency is doing, and determine for yourself if it is needed. In this case, `request-promise-native` is an es6 promise wrapper around `request`. Its purpose is to fetch the requested resource using HTTP GET. This is very basic functionality, and we should think about if we can replace this module with a better method, or a more modern dependency.

Because we are operating in a Chromium/Node environment (Electron), we can choose from several potential replacements. If we want to use node (which would allow us to move this code to the backend in the future), we could use `node-fetch`, `axios`, `got` or similar. However, since this code is running in the renderer thread of Electron, we have access to the chromium browser APIs, including `fetch` (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch). This is what I suggest you use here.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+  }, [localPath])
+
+  const getURL = uri =>
+  new Promise((resolve, reject) => {
+    try {
+      resolve(new URL(uri));
+    } catch (error) {
+      reject(error);
+    }
+  });
```
jthrilly, line 68: I would like you to think about what this code does, and then tell me why you think it is needed.

This is another example of you needing to be more critical about including code from other projects!

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+ */
+const downloadProtocolFromURI = inEnvironment((environment)=> {
+
+  if (environment === environments.ELECTRON) {
+    const request = require('request-promise-native');
+    const path = require('path');
+    const electron = require('electron');
+    const { dialog } = electron.remote
+    const tempPath = (electron.app || electron.remote.app).getPath('temp');
+    const from = path.join(tempPath, 'SampleProtocol') + '.netcanvas';
```
jthrilly, line 91: You should not hard code the name of the protocol here. This will cause a collision/overwrite and unpredictable results with the code is run multiple times.

You could either use a uuid, or you could use the protocol name. If you do the latter, you must handle the case where the temp directory path (including the filename) already exists.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+      dialog.showOpenDialog(null, {
+        properties: ['openDirectory']
+      })
+      .then((result) => {
+        if (result.filePaths.toString() !== ''){
+          const destination = path.join(result.filePaths.toString(), 'SampleProtocol') + '.netcanvas';
+          resolve(destination);
+        }
+      })
+    })
```
jthrilly, line 104: You create a problem for yourself here by using an open dialog rather than a save dialog.

Because you use an open dialog, you only get a path bath with no filename associated with it. This means that you once again incorrectly need to hard code the filename into the path. 

If you use a save dialog, you will also get a filename that the user has chosen. You can pre-populate this dialog with the actual sample protocol name, based on the URL.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+      
+    return (uri) => {
+      selectPath
+        .then((destination) => {
+          let promisedResponse;
+          promisedResponse = getURL(uri)
+            .catch(urlError)
+            .then(url => request({ method: 'GET', encoding: null, uri: url.href }));
```
jthrilly, line 112: Why is this necessary here?

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+
+          return promisedResponse
+            .catch(networkError)
+            .then(data => writeFile(from, data))
+            .catch(fileError)
+            .then(() => console.log(destination))
```
jthrilly, line 118: We should not use `console.log` in production unless absolutely necessary.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+            .then(() => {
+              fs.rename(from, destination, function(err){
+                if (err){
+                  throw error;
+                }
+                else {
+                  console.log("Successfully moved the file")
+                }
```
jthrilly, line 126: This is another small instance where you should use the return early pattern (see my other comment on this). Return if there is an error, and then you do not need an `else` statement.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+            .catch(urlError)
+            .then(url => request({ method: 'GET', encoding: null, uri: url.href }));
+
+          return promisedResponse
+            .catch(networkError)
+            .then(data => writeFile(from, data))
+            .catch(fileError)
+            .then(() => console.log(destination))
+            .then(() => {
+              fs.rename(from, destination, function(err){
```
jthrilly, line 120: Why not return `fs.rename` here, and then handle the error with `catch()`?

```
.then(() => fs.rename(from, destination))
.catch((err) => { // handle rename error })
```

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+                if (err){
+                  throw error;
+                }
+                else {
+                  console.log("Successfully moved the file")
+                }
+              })
+            })
+            .then(() => {
+              setLocalPath(destination)
```
jthrilly, line 130: Call `openNetcanvas` here, as discussed above.

src/components/Home/WelcomeHeader.js
```diff
@@ -37,6 +48,94 @@ const WelcomeHeader = () => {
...
+ * @param {string} uri
+ * @return {string} output filepath
+ */
+const downloadProtocolFromURI = inEnvironment((environment)=> {
+
+  if (environment === environments.ELECTRON) {
+    const request = require('request-promise-native');
+    const path = require('path');
+    const electron = require('electron');
+    const { dialog } = electron.remote
```
jthrilly, line 89: Because we do not need to worry about working in multiple environments, these `require` statements can be replaced with es6 module imports and moved to the top of the file for better consistency.

jthrilly (changes requested):
Thank you for this! Overall I am very pleased with your first attempt here. The functionality does work, and I can see you have tried hard to make it behave as it does in Interviewer. I also note your very good git commit messages :+1: . However, there are some quite important things that need to be changed before this is ready to merge. 

1. As you can see from the status checks on github, your code has lots and lots of linting errors. We use linting rules to enforce code style and best practices, and all code must lint before it can be merged. You need to install and use `eslint` in visual studio code so that these errors show up as you are coding. You can also run `npm run lint` from within the project, to get a list of current errors. Please find and fix all linting errors.
2. There is no feedback for the user about the process. The sample protocol is 25mb, so it is reasonable to expect it to take a while to download. The user must know what is happening during this period! In Interviewer we use the toast system to both communicate the progress, and allow the user to cancel if they wish. Architect has the concept of a 'lock' which can be used to block all interactions in the app until something has finished. My suggestion is that you use the toast system, because it is more interactive.
3. You need to structure the location of your code more logically. Complex logic does not belong in the same file as presentation code. In this case, the download methods should be in their own file in the utils folder.
4.  (small) I would prefer that the three buttons span the width of the welcome panel. The way you have it now causes the text to wrap, which I think looks quite ugly.

I've done a detailed review of the other important changes needed in the form of review comments. Please read them and let me know if you have any questions, either on slack or by replying to the comments directly.

package.json
```diff
@@ -268,5 +268,8 @@
             "@modules/(.*)$": "<rootDir>/src/ducks/modules/$1",
             "^react-native$": "react-native-web"
         }
+    },
+    "dependencies": {
+        "node-fetch": "^1.7.3"
```
jthrilly, line 273: Because we bundle our code using webpack, we don't use the `dependencies` section of the main package file. If the dependency is going to be bundled, add it to `devDependencies`, or alternatively if it is going to be used unbundled at runtime, add it to `public/package.json`.

src/components/Home/WelcomeHeader.js
```diff
@@ -217,11 +123,11 @@ const downloadProtocolFromURI = inEnvironment((environment)=> {
 };
 
 WelcomeHeader.PropTypes = {
-  openNetcanvas: PropTypes.func.isRequired,
-}
+  downloadProtocolFromURI: PropTypes.func.isRequired
```
jthrilly, line 126: Really pleased to see prop type validation!

jthrilly (approved):

jthrilly (dismissed):

PR #805, opened 2023-06-08
Fixes not able to close stage editor when filter disabled & validation that references variables

jthrilly:
### Disabling the "filter" from the node side panel removes the "save button" from the stage editor

Redux form still validates fields that are not shown, even when their value is null. We set fields to invisible with a value of null when they are contained within a toggleable section, in order to "disable" them.

Redux form has no concept of disabling a field.

In this case, the hidden field representing the node panel filter was still being validated, and the validation function was returning a string indicating that the field was empty (because we set it to null in order to reset it!). This error was hidden from the user, because the field was hidden and deregistered.

The fix is to change the validation function to return undefined if the value is null. Could also have been implemented as passing undefined as the validation prop to the field based on if the toggleable section is open or closed.

### Changing a variable type

This bug was caused by failing to correctly rest the validation value when the validation type was changed. It caused extremely strange (and invalid) behaviour when switching between types.

The fix here was to create a new function to calculate the "automatic" valid of validations that are created/changed to correctly set a value in `withUpdateHandlers`. This function attempts to keep the value if the type of the previous and current validation is the same, but otherwise resets it to null, which requires the user to create a new value.

### Variables that are only used as validation for other variables are still able to be deleted. This creates an invalid protocol

We did not check for variable use in the codebook, where variables can be referenced as validation targets when using `sameAs` or `differentFrom`. This meant that these variables could be deleted, causing an invalid protocol via a dangling reference. 

The fix was to include codebook variable paths in the search tree for variables, and to update the logic for calculating if a variable is in use (as well as identifying _where_ it is used) to account for these locations. This was a complicated fix!

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -18,6 +18,9 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '2.7.18'
+      # update apt cache
+      - name: Update apt cache
+        run: sudo apt-get update -y
       # Set node version
       - uses: actions/setup-node@v2
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -58,6 +58,9 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: '2.7.18'
+      # update apt cache
+      - name: Update apt cache
+        run: sudo apt-get update -y
       # Set node version
       - uses: actions/setup-node@v2
         with:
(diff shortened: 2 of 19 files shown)


buckhalt (approved):
Works for me. Verified that:

- Filters can be removed from the node side panel.
- Validation types can be correctly changed. Value is kept if the type of the previous and current validation is the same, but otherwise is reset and prompts/requires the user to add a new value. 
- Variables only used as validation for other variables cannot be deleted. Codebook correctly references which other variable the variable is being used as validation for.

PR #809, opened 2023-06-22
Fix/Skip logic Rule validation

buckhalt:
Skip logic section allowed skip logic to be configured with no rules. This caused the protocol to be invalid due to skip logic missing required property 'filter'. 

This fix ensures that skip logic is configured with at least one rule by marking the field as required.


buckhalt:
This branch also updates the interviewer reference to release 6.4.2 which fixes a bug where skip logic was not being evaluated properly. The bugfix is in networkQuery commit 60bd337.

jthrilly (approved):

PR #817, opened 2023-09-06
updated the readme file by adding a new step in Development section

mrkarimoff:
Updated the readme file by adding a new step of running `npm run preelectron:dev` in Development section and fixed the required NPM version to 8.3.2 to match the package.json file

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ For questions and support, please visit the [Network Canvas User Community](http
 ### Prerequisites
 
 - [Node.js](https://nodejs.org/en/) (v14.21.3)
-- [npm](https://www.npmjs.com/) (v8.3.2)
+- [npm](https://www.npmjs.com/) (v8.19.4)
 - [Git](https://git-scm.com/)
 - [Python](https://www.python.org/) (v3.10.12)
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,20 +45,20 @@ Note: for Apple Silicon users, you need to install the `electron` package manual
...
+| `test`                          | Runs testing suite                                                                                                             |
+| `preflight`                     | Runs linting & testing. Useful as a prepush/build hook                                                                         |
+| `dist:mac`                      | Build and publish OS X verison                                                                                                 |
+| `dist:linux`                    | Build and publish Linux version                                                                                                |
+| `dist:win`                      | Build and publish Windows version                                                                                              |
+| `dist:all`                      | Build and publish all platforms                                                                                                |
+| `update-submodules`             | Update git submodules                                                                                                          |
```
jthrilly, line 61: Let's add the `preelectron:dev` command here, for reference.

README.md
```diff
@@ -13,7 +13,7 @@ For questions and support, please visit the [Network Canvas User Community](http
 ### Prerequisites
 
 - [Node.js](https://nodejs.org/en/) (v14.21.3)
-- [npm](https://www.npmjs.com/) (v8.19.4)
+- [npm](https://www.npmjs.com/) (v8.3.2)
```
jthrilly, line 16: I think this was correct previously - if package.json says 8.3.2 I would suggest updating that, instead.

README.md
```diff
@@ -75,13 +75,18 @@ e.g.
...
+
 1. `npm run start:network-canvas:electron`: to start the webpack dev server
-  - Note: must be running on port 3000.
-2. `npm run electron:dev` (in another terminal session)
-  1. Copies the electron source to `./electron-dev`
-  2. Runs the electron app from there
 
+- Note: must be running on port 3000.
+
+2. `npm run preelectron:dev` (in another terminal session)
```
jthrilly, line 85: I think we should make it clear that this step only needs to happen when you first set up the repo, or when you bump the version number.

jthrilly (changes requested):

README.md
mrkarimoff: Sure, will do sir

jthrilly (approved):

PR #840, opened 2024-12-04
Feature/improve filter messaging

buckhalt:
Edge stage-level filters can sometimes lead to unexpected behavior when configured with edge creation or edge display on a sociogram. 

For instance:

> Filter: "edges of type `a` must exist"
> Edge creation: create edge type `b`

In this case, edges of type `b` will not be shown on the sociogram.

This PR adds warning tips to the edge creation configuration, edge display configuration, and filter configuration sections when there is a potential mismatch between the filter and the edges to be created or displayed.


src/components/Query/Rules/withEntityRuleType.js
```diff
@@ -10,9 +10,9 @@ const entityRuleTypes = {
   TYPE_RULE,
 };
 
-const entityRuleTypeOptions = [
-  { label: 'Attribute - rule based on the value of this alter type\'s attributes.', value: VARIABLE_RULE },
-  { label: 'Presence - based on the presence or absence of this alter type in the interview network.', value: TYPE_RULE },
+const entityRuleTypeOptions = (entityType) => [
+  { label: `Attribute - rule based on the value of this ${entityType} type's attributes.`, value: VARIABLE_RULE },
+  { label: `Presence - based on the presence or absence of this ${entityType} type in the interview network.`, value: TYPE_RULE },
```
jthrilly, line 15: Nice improvement!

src/components/sections/Filter.js
```diff
@@ -30,6 +32,27 @@ const Filter = () => {
...
+      .flatMap((prompt) => prompt.edges.display);
+
+    if (edgeCreationValues.length > 0 || edgeDisplayValues.length > 0) {
+      shouldShowWarning = getEdgeFilteringWarning(
+        currentValue.rules,
+        [...edgeCreationValues, ...edgeDisplayValues],
+      );
+    }
+  }
+
```
jthrilly: While this works, there are a few rough edges with it:
- The `edgeCreationValues` and `edgeDisplayValues` variables (derived state) aren't memoized, so they will be recomputed on every render. This isn't terrible in this situation because the computation is relatively cheap, but we can explicitly tell React when the values need to be recomputed using `useMemo`. Note that this will possibly be automatic once we can use the React compiled in React 19.
- Similarly, the way you have set up `shouldShowWarning` isn't a particularly "React-y" way of doing things. Try to think of it in terms of "a computed value that must update under certain circumstances", rather than the way you have it now which is more of a vanilla JS way of thinking about it as "a variable that I will update the value of". Not quite sure I am explaining this well, but what this comes back to is the often quoted Dan Abramov line `ui = fn(data)(state)`. 

So you could address those issues like this:

```suggestion
  const prompts = useSelector((state) => getFormValue(state, "prompts"));
  
  const { edgeCreationValues, edgeDisplayValues } = useMemo(() => {
    if (!prompts) return { edgeCreationValues: [], edgeDisplayValues: [] };
    const edgeCreationValues = [];
    const edgeDisplayValues = [];

    prompts.forEach((prompt) => {
      if (prompt?.edges?.create) edgeCreationValues.push(prompt.edges.create);
      if (prompt?.edges?.display) edgeDisplayValues.push(...prompt.edges.display);
    });

    return { edgeCreationValues, edgeDisplayValues };
  }, [prompts]);

  const shouldShowWarning = useMemo(() => {
    if (edgeCreationValues.length > 0 || edgeDisplayValues.length > 0) {
      return getEdgeFilteringWarning(currentValue.rules, [...edgeCreationValues, ...edgeDisplayValues]);
    }
    return false;
  }, [currentValue, edgeCreationValues, edgeDisplayValues]);
```

An alternative would be to make a selector that encapsulates all of this:

```
const getFormValue = formValueSelector('edit-stage');
const getPrompts = (state) => getFormValue(state, 'prompts');
const getCurrentValue = (state) => getFormValue(state, 'filter');

export const getShouldShowEdgeWarning = createSelector(
  [getPrompts, getCurrentValue],
  (prompts, currentValue) => {
    if (!prompts) return { edgeCreationValues: [], edgeDisplayValues: [], shouldShowWarning: false };

    const edgeCreationValues = [];
    const edgeDisplayValues = [];

    prompts.forEach((prompt) => {
      if (prompt?.edges?.create) edgeCreationValues.push(prompt.edges.create);
      if (prompt?.edges?.display) edgeDisplayValues.push(...prompt.edges.display);
    });

    const shouldShowWarning = (edgeCreationValues.length > 0 || edgeDisplayValues.length > 0) 
      ? getEdgeFilteringWarning(currentValue.rules, [...edgeCreationValues, ...edgeDisplayValues])
      : false;

    return shouldShowWarning;
  }
);
```

I think conceptually speaking, it is less readable to derive values inside a selector like this. If you wanted to encapsulate the logic in a single place, a custom hook would probably be more appropriate. However, since this logic isn't reused, it doesn't really make sense to go to that trouble.

src/components/sections/Filter.js
```diff
@@ -54,13 +77,21 @@ const Filter = () => {
...
+          You can optionally filter which nodes or edges are shown on this stage, by creating
           one or more rules using the options below.
         </p>
       )}
       startExpanded={!!currentValue}
       handleToggleChange={handleToggleChange}
     >
+      {shouldShowWarning && (
+      <Tip type="warning">
+        <p>
```
jthrilly, line 91: Just a note to please check that the `Tip` component is typically rendered with a paragraph inside of it, if you haven't already :)

src/components/sections/SociogramPrompts/PromptFieldsEdges.js
```diff
@@ -7,7 +7,7 @@ import { change, Field, formValueSelector } from 'redux-form';
 import * as Fields from '@codaco/ui/lib/components/Fields';
 import { Section, Row } from '@components/EditorLayout';
 import Tip from '../../Tip';
-import { getEdgesForSubject } from './selectors';
+import { getEdgeFilteringWarning, getEdgeFilters, getEdgesForSubject } from './selectors';
```
jthrilly: I'm reading this chronologically, but it is suspicious to me that `getEdgeFilterWarning` is being imported from `./selectors` but then isn't used as a selector (via `useSelector`)... 🤔

src/components/sections/SociogramPrompts/PromptFieldsEdges.js
```diff
@@ -59,6 +62,15 @@ const DisplayEdges = ({ form, entity, type }) => {
         }}
       >
         <Row>
+          {shouldShowNetworkFilterWarning && (
+          <Tip type="warning">
+            <p>
+              One or more of the selected edge types are not currently included in the
+              stage-level network filtering. If they are not included, the edges will not be
+              displayed.
```
jthrilly:
```suggestion
              Stage level network filtering is enabled, but one or more of the edge types 
              you have configured to display on this prompt are not currently included in the 
              filter. This means that these edges may not be displayed. Either remove the 
              stage-level network filtering, or add these edge types to the filter to resolve this 
              issue.
```

src/components/sections/SociogramPrompts/PromptFieldsTapBehaviour.js
```diff
@@ -168,13 +174,24 @@ const TapBehaviour = ({
           />
         )}
         { tapBehaviour === TAP_BEHAVIOURS.CREATE_EDGES && (
+        <>
+          {showNetworkFilterWarning && (
+          <Tip type="warning">
+            <p>
+              The selected edge type is not currently included in the stage-level network filtering.
+              If it is not included, the edge will not be displayed when it is created.
```
jthrilly:
```suggestion
              Stage level network filtering is enabled, but the edge type you want to create 
              on this prompt is not currently included in the filter. This means that these 
              edges may not be displayed. Either remove the stage-level network filtering, 
              or add these edge types to the filter to resolve this issue.
```

src/components/sections/SociogramPrompts/PromptFieldsTapBehaviour.js
```diff
@@ -101,6 +102,11 @@ const TapBehaviour = ({
     return true;
   };
 
+  const selectedValue = useSelector((state) => getFormValue(state, 'edges.create'));
+
+  const edgeFilters = useSelector((state) => getEdgeFilters(state));
+  const showNetworkFilterWarning = getEdgeFilteringWarning(edgeFilters, [selectedValue]);
```
jthrilly:
```suggestion
  const showNetworkFilterWarning = useSelector(getShouldShowEdgeFilterWarning)
```

(see below for implementation)

As an aside, when you have `useSelector((state) => getEdgeFilters(state))` or any situation where the parameters of the inner and outer function are the same (or compatible), you can just do `outerFn(innerFn)` and the parameters will be automatically passed. So in this case `useSelector(getEdgeFilters)`.

src/components/sections/SociogramPrompts/selectors.js
```diff
@@ -32,3 +33,57 @@ export const getEdgesForSubject = (state) => {
...
+
+    // case 3
+    if (isEdgeInDoesNotExistFilters) {
+      return true;
+    }
+
+    // No warning in other cases
+    return false;
+  });
+};
```
jthrilly: Yeah, so as I thought from the way it is used, this is not quite right. `getFilterEdgeWarning` isn't a selector. It computes a value based on two _other_ selectors. So it is just a normal function that should live elsewhere. 

You can handle this a few ways, but what I would probably do is to combine the logic into an _actual_ selector with something like this (not tested): 

```suggestion
export const getShouldShowEdgeFilterWarning = (state) => {
  const getStageValue = formValueSelector('edit-stage');
  const currentFilters = getStageValue(state, 'filter');
  const edges = getFormValue(state, 'edges.create');

  if (!currentFilters || !currentFilters.rules) {
    return false;
  }

  const edgeFilters = currentFilters.rules.filter((rule) => rule.type === 'edge');
  const existFilters = edgeFilters.filter((rule) => rule.options.operator === 'EXISTS');
  const doesNotExistFilters = edgeFilters.filter((rule) => rule.options.operator === 'NOT_EXISTS');

  return edges.some((edge) => {
    const isEdgeInExistFilters = existFilters.some((rule) => rule.options.type === edge);
    const isEdgeInDoesNotExistFilters = doesNotExistFilters.some(
      (rule) => rule.options.type === edge,
    );

    if (isEdgeInExistFilters) {
      return false;
    }

    if (!isEdgeInExistFilters && existFilters.length > 0) {
      return true;
    }

    if (isEdgeInDoesNotExistFilters) {
      return true;
    }

    return false;
  });
};
```

Alternatively, you could turn `getEdgeFilteringWarning` into a selector by using `createSelector`, and having it fetch its dependencies (filters and edges) from state using those other selectors:

```
export const getEdgeFilteringWarning = createSelector(
getEdgeFilters,
(state) => getFormValue(state, 'edges.create'), // Could define this as its own selector, `getPromptEdgeCreate`
(filters, edges) => {
  const existFilters = filters.filter((rule) => rule.options.operator === 'EXISTS');
  const doesNotExistFilters = filters.filter((rule) => rule.options.operator === 'NOT_EXISTS');

  // if any edge should show a warning, return true
  return edges.some((edge) => {
    const isEdgeInExistFilters = existFilters.some((rule) => rule.options.type === edge);
    const isEdgeInDoesNotExistFilters = doesNotExistFilters.some(
      (rule) => rule.options.type === edge,
    );

    // case 1
    if (isEdgeInExistFilters) {
      return false;
    }

    // case 2
    if (!isEdgeInExistFilters && existFilters.length > 0) {
      return true;
    }

    // case 3
    if (isEdgeInDoesNotExistFilters) {
      return true;
    }

    // No warning in other cases
    return false;
  });
});
```

jthrilly (commented):
Some ideas here for you. Maybe you want to review the docs on redux a bit? Let me know if you want to chat about any of the concepts. This is something I had no idea about for 2+ years, so don't worry if it is confusing!

src/components/sections/Filter.js
buckhalt: Double checked that it is typically rendered with a paragraph inside of it :)

src/components/sections/SociogramPrompts/PromptFieldsTapBehaviour.js
buckhalt: resolved https://github.com/complexdatacollective/Architect/pull/840/commits/55b9b72d7efe2cffd8baaee48478867d16da5490

src/components/sections/SociogramPrompts/PromptFieldsEdges.js
buckhalt: resolved https://github.com/complexdatacollective/Architect/pull/840/commits/55b9b72d7efe2cffd8baaee48478867d16da5490

src/components/sections/SociogramPrompts/PromptFieldsTapBehaviour.js
buckhalt: resolved [3422a77](https://github.com/complexdatacollective/Architect/pull/840/commits/3422a77b6b5bc3fa4cfd493966982cf82585eafa)

src/components/sections/Filter.js
buckhalt: resolved with first approach [3422a77](https://github.com/complexdatacollective/Architect/pull/840/commits/3422a77b6b5bc3fa4cfd493966982cf82585eafa)

src/components/sections/SociogramPrompts/PromptFieldsEdges.js
buckhalt: resolved [3422a77](https://github.com/complexdatacollective/Architect/pull/840/commits/3422a77b6b5bc3fa4cfd493966982cf82585eafa)

src/components/sections/SociogramPrompts/selectors.js
buckhalt: Because each use case of this function needs to get edges and filters in slightly different ways, I'm not sure there's an easy way to make this an actual selector.  I think the most straightforward implementation is to leave the function signature the same but treat it as a utility function.

Moved to a `utils` file and treated as a utility func instead of selector [3422a77](https://github.com/complexdatacollective/Architect/pull/840/commits/3422a77b6b5bc3fa4cfd493966982cf82585eafa)

